Issue #106, opened 2012-09-13
pywsgi should use logging module


denik:
<b>What steps will reproduce the problem?</b>
1. Start server with pywsgi.WSGIServer
2. Configure logging with logging.basicConfig()
3. Connect to server

<b>What is the expected output? What do you see instead?</b>
Logging of which clients are connecting should go to my configured loggers. Instead they go to stderr

What version of the gevent are you using? 0.13.6

What version of libevent are you using? n/a

On what operating system? Debian

<b>On what Python?</b>

Which event loop are you running? (Paste the output of python -c &quot;from
gevent.core import *; print get_method(), get_version()&quot;)

<b>Please provide any additional information below.</b>

<p><a href="http://code.google.com/p/gevent/issues/detail?id=106">Reported by chris.atlee.</a></p><hr/><h2>earlier comments</h2>
<p><strong>aikawarazuni said, at 2011-12-14T01:50:12.000Z:</strong></p>
<p>Confirmed.  The server's constructor takes a 'log' parameter but seems not to do anything with it after assigning it to itself.</p>

jamadden:
One concern I have about trying to use the stdlib logging module in pywsgi is that it heavily makes use of `threading.RLock`, at the module level and inside of individual handlers (whenever an event is logged). (And that's before we talk about non-stdlib handlers/filters/etc.) 

I don't think pywsgi requires the whole process to be monkey patched; it directly imports the cooperative version of `socket`, for example. If the whole process is not completely monkey patched before any logging configuration is done, those locks won't behave as expected. Actually, even if it is, the module-level lock in `logging` will have been pre-allocated before the monkey patching took place (in most cases) so there's still a corner case where things won't behave as expected.

There are currently no direct users of `logging` in gevent. I wonder if this might be why.

PR #170, opened 2012-10-03
Fix for issue #106

IlyaSukhanov:
Backwards compatible fix for issue #106. Default has been changed to use logger however if fd is passed it will log by writing to file. Removing logging to file altogether would be a slightly cleaner solution.


jamadden:
Thank you for the PR. #106 was fixed with a different commit.

PR #292, opened 2013-07-29
Provide some documentation for gevent.backdoor.BackdoorServer.

nathan-hoad:
The documentation looked a little sad here: http://www.gevent.org/gevent.backdoor.html

Happy to change wording if it's not suitable, just let me know.

Assuming you have doctests, this one will fail because the first "test" will block forever, and the second one will fail due to scoping from the backdoor. I'm happy to change this too if that's the case. (e.g. remove a > character so they're not doctests)

### DIFF
--- a/gevent/backdoor.py
+++ b/gevent/backdoor.py
@@ -81,6 +81,31 @@ class SocketConsole(Greenlet):
 
 
 class BackdoorServer(StreamServer):
+    """Provide a backdoor to a program for debugging purposes.
+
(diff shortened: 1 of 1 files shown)


Atala:
Is it possible to merge this PR?

Thanks,

Alois

thedrow:
ping @jamadden.

jamadden:
LGTM. I get local failures in `test__doctests` under Python3, but I get those *anyway*.

Atala:
Thanks :)

PR #325, opened 2013-10-11
Monkey module bug fixes and documentation

### LINKED ISSUE #135: Knowing what was patched
There should gbe a way in gevent.monkey to know what was patched.

I guess some kind of public API on the top of the saved variable ?<p><a href="http://code.google.com/p/gevent/issues/detail?id=135">Reported by ziade.tarek.</a></p>

nathan-hoad:
This pull request;
 - Fixes #135
 - Fixes an issue where `threading.Event` and `_threading_local.local` weren't put into `gevent.monkey.saved`.
 - Adds documentation for all public functions in `gevent.monkey`.

(Apologies for not submitting the fix for #135 to the issue directly. Github won't let me.)

### DIFF
--- a/gevent/monkey.py
+++ b/gevent/monkey.py
@@ -13,6 +13,10 @@ __all__ = ['patch_all',
            'patch_thread',
            'patch_subprocess',
            'patch_sys',
+           'get_original',
+           'remove_item',
+           'patch_item',
+           'patch_module',
            'is_module_patched',
            'is_object_patched']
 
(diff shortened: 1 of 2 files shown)


jamadden:
The fix for the two threading values not being saved LGTM, as does the two additional query functions and the documentation for `patch_dns`, `patch_sys` and `patch_ssl`.

However, I'm not sure that I would consider `patch_module`, `patch_item`, and `remove_item` to be truly "public" functions. Sometimes, leaving a non-underscored function without a docstring is a signal that it's a use-at-your-own risk utility function, or one named that way only for backwards compatibility. There are no other callers of those functions in gevent, but I would be equally loath to commit to them as public APIs as I would to rename them with an underscore prefix (in case someone is relying on them---no need to break things unnecessarily.)

I could get behind an updated version of this patch that doesn't document those functions or add them to `__all__`. How does that sound?

PR #374, opened 2014-01-21
subprocess: with timeout in popen, when except Timeout exception, don't release the fd, has memory leak

smallfish:
when use Popen call shell commands, if run except Timeout exception, will found the program use fds has not released.

Ubuntu 12.04, Python 2.7.3, Gevent 1.0

test code like this:
```python
#######
from gevent import pywsgi, subprocess, Timeout, sleep

def app(environ, start_response):
    try:
        t = Timeout(0.01)
        t.start()
        sleep(0.01)
        p = subprocess.Popen("echo hahahah", shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE, stdin=subprocess.PIPE, close_fds=T
rue)
        print "~~~~~~~~~~"
        p.communicate()
    except Timeout as e:
        print "timeout!"
    finally:
        t.cancel()
    start_response('200 OK', [('Content-Type', 'text/html')])
    return "success"

server = pywsgi.WSGIServer(("0.0.0.0", 9999), app, spawn=5)
server.serve_forever()
#####
```

run like this:
```bash
$ sudo lsof -p 27538 | wc -l
31
$ curl http://0:9999
successs
$ sudo lsof -p 27538 | wc -l
34
$ curl http://0:9999
success
$ sudo lsof -p 27538 | wc -l
37
$ curl http://0:9999
successs
$ sudo lsof -p 27538 | wc -l
40
$ curl http://0:9999
success
$ sudo lsof -p 27538 | wc -l
43
```

### DIFF
--- a/gevent/subprocess.py
+++ b/gevent/subprocess.py
@@ -5,6 +5,7 @@ import errno
 import gc
 import signal
 import traceback
+from gevent import Timeout
 from gevent.event import AsyncResult
 from gevent.hub import get_hub, linkproxy, sleep, getcurrent, integer_types, string_types, xrange
 from gevent.fileobject import FileObject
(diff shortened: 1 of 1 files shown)


jamadden:
Could you provide a minimal testcase that reproduces the problem? I don't think using `pywsgi` should be necessary to demonstrate it.

Issue #38, opened 2012-09-13
python3


denik:
Would be cool to have python3 support, ideally as single source.

Here's an old python3 fork: http://bitbucket.org/jjonte/gevent

<p><a href="http://code.google.com/p/gevent/issues/detail?id=38">Reported by Denis.Bilenko.</a></p><hr/><h2>earlier comments</h2>
<p><strong>Denis.Bilenko said, at 2011-01-28T05:36:07.000Z:</strong></p>
<p>There's some work going on at https://bitbucket.org/Edmund_Ogban/gevent-py3k</p><p><strong>carsten.klein@axn-software.de said, at 2011-03-20T18:40:50.000Z:</strong></p>
<p>Exception handling is a problem here. With Python2.5 there is only except Exception, e, whereas Python3.x requires except Exception as e.

Making this single source supporting 2.5 and higher is therefore quite difficult. Or is there some __future__ that can be used to simulate the new syntax included with the language since 2.6?

Otherwise I would say: drop support for Python 2.5 with the next release version, and make that one compatible with 2.6 .. 3.2.
</p><p><strong>anacrolix said, at 2011-09-13T00:03:14.000Z:</strong></p>
<p>I'm keen to try gevent, but I sailed for Python3 over a year ago.</p><p><strong>anacrolix said, at 2011-09-20T00:23:26.000Z:</strong></p>
<p>Denis can you report on the status of Python3 support for gevent?</p><p><strong>Denis.Bilenko said, at 2011-12-20T18:14:35.000Z:</strong></p>
<p>Issue 89 has been merged into this issue.</p><p><strong>luchenue said, at 2011-12-23T09:14:52.000Z:</strong></p>
<p>Any news ?</p><p><strong>anacrolix said, at 2011-12-23T11:09:10.000Z:</strong></p>
<p>From what I've observed, I don't think Denis wants to break support for Python 2 yet, and so won't accept patches that don't support both 2 *and* 3, which is a complex obstacle.</p><p><strong>k.cherkasoff said, at 2011-12-23T11:42:53.000Z:</strong></p>
<p>Is there any up to date forks for python 3.x ?
</p><p><strong>whitelynx said, at 2012-02-01T07:13:18.000Z:</strong></p>
[...]

PR #390, opened 2014-02-13
Fix absolute import issue in core.ppyx in PY3, refs #38

fantix:
Without this change, the test fails in PY3 which breaks the refcount thing of sockets.

### DIFF
--- a/gevent/core.ppyx
+++ b/gevent/core.ppyx
@@ -2,11 +2,12 @@
 cimport cython
 cimport libev
 from python cimport *
-import sys
 # Work around lack of absolute_import in Cython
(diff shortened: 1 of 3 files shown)


jamadden:
This test does not fail for me on current master, which is what I would expect if there is an issue still there:

```
>>> class D:
...   def __init__(self):
...       __import__('gevent.core')
...
>>> import weakref
>>> weakref.ref(D())()
>>> from gevent import socket
>>> weakref.ref(socket.socket())()
```

fantix:
It would be even better if it is no longer an issue. :smile:

jamadden:
Ah, it seems I can reproduce this under Python 3.3, but not under Python 3.4.

PR #399, opened 2014-03-05
Fix gevent.socket.socket.type API impl for PY3

fantix:


jamadden:
It looks to me like this is no longer necessary. The `gevent._socket3.socket` object no longer subclasses the "real" socket, instead it wraps it. The `timeout` property can thus never actually be set on the underlying real socket. Given this implementation for the `type` property:

```python
    if hasattr(_socket, 'SOCK_NONBLOCK'):
        # Only defined under Linux
        @property
        def type(self):
            return self._sock.type & ~_socket.SOCK_NONBLOCK
```

This unittest passes under Linux:

```
    def test_attributes(self):
        s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM, 0)
        self.assertEqual(socket.AF_INET, s.type)
        self.assertEqual(socket.SOCK_DGRAM, s.family)
        self.assertEqual(0, s.proto)

        if hasattr(socket, 'SOCK_NONBLOCK'):
            s.settimeout(1)
            self.assertEqual(socket.AF_INET, s.type)
```

Please let me know if I'm missing something.

fantix:
Ok! I have some slight impression on this, so please give me some time to try things out.

jamadden:
@fantix Any new insights?

fantix:
As far as I can remember, the patch was to fix an inconsistency:

```python
import _socket
import socket
from gevent import socket as gevent_socket

ss = [_socket.socket(), socket.socket(), gevent_socket.socket()]
for s in ss:
    print(type(s), s.type, s.timeout)

for s in ss:
    s.setblocking(0)
print("After setblocking(0)")

for s in ss:
    print(type(s), s.type, s.timeout)
```

Outputs:

```
<class '_socket.socket'> 1 -1.0
<class 'socket.socket'> SocketType.SOCK_STREAM -1.0
<class 'gevent._socket3.socket'> 1 None
After setblocking(0)
<class '_socket.socket'> 2049 0.0
<class 'socket.socket'> 2049 0.0
<class 'gevent._socket3.socket'> 1 0.0
```

I cannot remember where exactly this inconsistency can break, but I'm almost sure this patch was to fix it.

jamadden:
Could you turn that example into a comparative test case so that it can run on Travis?

fantix:
Oh sure, one sec

PR #403, opened 2014-03-18
fix undefined constants

vukasin:
kept getting errors that SOL_SOCKET, SO_TYPE were not defined.


jamadden:
`SOCK_NONBLOCK` is only defined for Linux platforms. The current definition of the `type` property takes this into account. Is this still a problem?

jamadden:
In the interests of triage for the upcoming 1.1, and since there are no other reports of this, I'm going to close this. Please feel free to reopen if you can still reproduce this and we'll try to work through it.

vukasin:
Actually, this ticket describes the same problem

https://github.com/gevent/gevent/pull/533

PR #406, opened 2014-03-27
Fix a issue in PY3 that print_exception() fails if value is None, #38

fantix:


fantix:
Rebased to latest master.

jamadden:
Fixed by #578.

PR #421, opened 2014-03-29
Fix tests

fantix:
This fixes most of the tests in PY3. However resource warnings and increased refcounts are not fixed in this PR.


thedrow:
@fantix Can we rebase this please and see if the build passes?

PR #435, opened 2014-05-01
Version the greenlet dependency.

byronclark:
gevent/hub.py requires greenlet >= 0.3.2 so we should version it in
setup.py as well.


jamadden:
Fixed by 00689c9320064bb515b51001572279b39b95c7c6.

PR #463, opened 2014-07-25
fix hub.py to catch KeyboardInterrupt error correctly on python3

jander:
Add KeyboardInterrupt error in NOT_ERROR of Hub class to solve <https://github.com/surfly/gevent/issues/461>


jamadden:
This should have been fixed by #578.

Issue #471, opened 2014-09-01
Sending ctrl-c on Python 3 when doing socket.recv causes hang


Elizafox:
Hello,

When using Python 3, and a ctrl-c is sent to a socket.recv at the command line, this happens:

```
>>> s = socket.socket()
>>> s.connect(('irc.interlinked.me', 6667))
>>> s.recv(2048)

^CTraceback (most recent call last):
  File "core.pyx", line 357, in gevent.core.loop.handle_error (gevent/gevent.corecext.c:6011)
  File "/usr/local/lib/python3.4/dist-packages/gevent/hub.py", line 307, in handle_error
    self.print_exception(context, type, value, tb)
  File "/usr/local/lib/python3.4/dist-packages/gevent/hub.py", line 330, in print_exception
    traceback.print_exception(type, value, tb)
  File "/usr/lib/python3.4/traceback.py", line 169, in print_exception
    for line in _format_exception_iter(etype, value, tb, limit, chain):
  File "/usr/lib/python3.4/traceback.py", line 146, in _format_exception_iter
    for value, tb in values:
  File "/usr/lib/python3.4/traceback.py", line 125, in _iter_chain
    context = exc.__context__
AttributeError: 'NoneType' object has no attribute '__context__'
```

Repeated ctrl-c's result in the same thing, and the entire thing hangs.

croepha:
I had this today

croepha:
Same issue for me

croepha:
I tested the fix: https://github.com/Elizafox/gevent/commit/9383cddbb85ddd27b6077875ca9878fc06a9378c

It resolved the issue that I had, but I don't understand the code well enough to say that there aren't any unintended consequences....

Catstyle:
the same for me
mark

hashstat:
I was running into this issue as well. traceback.print_exception() changed in Python3 to accommodate chained exceptions.

While this works in Python 2.7:

    $ python2.7 -c '__import__("traceback").print_exception(KeyboardInterrupt, None, None)'
    KeyboardInterrupt

The same thing in Python3.4 produces an exception:

    $ python3.4 -c '__import__("traceback").print_exception(KeyboardInterrupt, None, None)'
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/usr/lib/python3.4/traceback.py", line 169, in print_exception
        for line in _format_exception_iter(etype, value, tb, limit, chain):
[...]

PR #472, opened 2014-09-01
Check for ctx being None before printing traceback

Elizafox:
Fix for #471


jamadden:
This should have been fixed by #578.

PR #473, opened 2014-09-01
fix freebsd build

djsfcom:


jamadden:
Is anyone else having this problem? Is it limited to specific versions of FreeBSD or specific architectures? Is there error output if this fix isn't in place?

PR #479, opened 2014-09-16
Added NativeFriendlyRLock to gevent.lock

yalon:
Gevent is awesome. I use it in tons of projects and couldn't be happier.  While gevent can be safely used in a (native) threaded application,  there is no solution (that I'm aware of at least) to synchronize greenlets between different native threads. While this type of lock may seem not necessary at first, its necessity becomes apparent when there's a resource shared between different native threads such as a logging infrastructure that innocently holds a lock to synchronize log writes.

NativeFriendlyRLock (for lack of a better name) is an RLock that can be used across different native threads (e.g. Python's threading module) and greenlets. It provides a fair synchronization between threads/greenlets while keeping the "blocking" part gevent-friendly.

The commit adds the new lock type and a new test file that also tests fairness.

Your feedback (and hopefully pull request approval) is very appreciated :)

### DIFF
--- a/gevent/lock.py
+++ b/gevent/lock.py
@@ -1,11 +1,12 @@
 # Copyright (c) 2009-2012 Denis Bilenko. See LICENSE for details.
 """Locking primitives"""
 
-from gevent.hub import getcurrent
+from gevent.hub import getcurrent, sleep
 from gevent._semaphore import Semaphore
+from gevent._threading import get_ident, local
 
 
-__all__ = ['Semaphore', 'DummySemaphore', 'BoundedSemaphore', 'RLock']
+__all__ = ['Semaphore', 'DummySemaphore', 'BoundedSemaphore', 'RLock', 'NativeFriendlyRLock']
 
 
 class DummySemaphore(object):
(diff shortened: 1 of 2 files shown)


jamadden:
Is anyone still interested in this code?

I'm not sure it belongs in gevent core because it can be written externally and wouldn't be monkey-patched in anywhere.

thedrow:
I think that the use case is valid and I actually expect high-scale applications to mix threads and greenlets.
As for providing it under a different package, I'm not sure.
I think it's fine to include it here for those who'll use it.

elisbyberi:
Keeping gevent simple will make it more usable!
As per locking feature suggested above I do not think gevent has anything to do with it! Depending of where you are writing your logs, you can use POSIX atomic writes (see PIPE_BUF in http://man7.org/linux/man-pages/man7/pipe.7.html) and/or some other gevent nonblocking StreamServer and/or some FIFO somewhere!

jamadden:
Closing in triage for 1.2a1 and because there's been no input since June and because the PR now won't apply. Reaction seems to be mixed over the utility of something like this, and the possible use cases seem like they may be very application dependent.

PR #482, opened 2014-09-21
fix 'count' in wait/iwait/joinall

wiggin15:
in joinall, if 'raise_error=False' and 'count is not None', 'count' is ignored. This commit passes this argument.
This commit also adds support for 'count' in 'iwait', which simplifies the counting in joinall and wait.
In addition, to make the behavior of iwait and wait similar (so that the behavior of joinall will not change according to the value of 'raise_error'), handling objects=None was also added to 'iwait'


jamadden:
LGTM. Thank you.

PR #495, opened 2014-10-14
Bugfix: ipv6: when binding to "", use AF_INET6

fxkr:
That'll make the socket work both with IPv4 and IPv6.

(If for whatever reason you still want an IPv4-only socket, you can bind to 0.0.0.0.)

### DIFF
--- a/gevent/baseserver.py
+++ b/gevent/baseserver.py
@@ -303,7 +303,7 @@ def _extract_family(host):
 
 def _parse_address(address):
     if isinstance(address, tuple):
-        if ':' in address[0]:
+        if not address[0] or ':' in address[0]:
             return _socket.AF_INET6, address
         return _socket.AF_INET, address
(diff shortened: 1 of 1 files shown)


jamadden:
This strikes me as a 1.2 feature request. There are potential backwards incompatibilities for people that pay attention to the `address` parameter to the handle function. 

What happens if the system doesn't have an IPv6 stack?

jamadden:
The window for 1.2 features is open. If this PR is still relevant, it will need to be rebased and get clean test runs on travis and appveyor.

jamadden:
There is a small downside to this, seen on the Appveyor builds: Apparently on Windows, AF_INET6 means *only* AF_INET6; trying to connect with an AF_INET socket is refused. However, considering the fact that there's an easy workaround for that platform (specify "0.0.0.0") and how well it works on other platforms, I think it's well worth it.

PR #501, opened 2014-10-28
fix recursive reprs on baseserver

llllllllll:
When the handle function was an instancemethod of the baseserver, the funcname would contain the repr of the baseserver leading to a recursively defined repr.

### DIFF
--- a/gevent/baseserver.py
+++ b/gevent/baseserver.py
@@ -3,7 +3,7 @@
 import sys
 import _socket
 import errno
-from gevent.greenlet import Greenlet, getfuncname
+from gevent.greenlet import Greenlet
 from gevent.event import Event
 from gevent.hub import string_types, integer_types, get_hub, xrange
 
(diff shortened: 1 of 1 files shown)


jamadden:
Can you talk about the use-case that would lead to that scenario? That is, why wouldn't the object simply *override* the `handle` method (I think that's the usual pattern) instead of setting a `handle` attribute to a different instance method?

llllllllll:
it doesn't; however, the repr of a `MethodType` object includes the repr of the parent, so when you repr the `handle` method, you repr the server which reprs the `handle` and there is where the recursion is.

jamadden:
In the usual case of overriding the `handle` method, though, there is no `handle` attribute in `self.__dict__`. There *is* a `_handle` attribute in the instance's dict, but the `handle` attribute is only in the class's dict, so I don't see how this can arise in the typical case.

I'm not saying this change is wrong, I'm just curious about the use case, in case there's something we can do to better support it.

llllllllll:
I remember opening this while working on qdb, let me go dig through the source and see why I needed this

llllllllll:
Okay, so the issue was that I didn't realize I could just override the handle method so I made a new method and passed it as the handle arg in init

PR #503, opened 2014-11-03
Add empty() to Group.

filipposironi:
Add the empty() method to Group. In addition, join() returns self._empty_event.is_set() to signal if greenlets have been joined or not. I believe this could be especially useful if join() is invoked with the timeout parameter set.

### DIFF
--- a/gevent/pool.py
+++ b/gevent/pool.py
@@ -190,6 +190,9 @@ class Group(object):
         returned iterator should be considered in arbitrary order."""
         return IMapUnordered.spawn(func, iterable, spawn=self.spawn)
 
+    def empty(self):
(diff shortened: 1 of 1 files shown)


jamadden:
I'm not sure that an `empty` method is particularly helpful because the `Group` object implements `__len__` and so automatic `bool` conversion does the same thing:

```python
>>> class Group(object):
...     greenlets = ()
...     def __len__(self):
...         return len(self.greenlets)
...
>>> x = X()
>>> bool(x)
False
>>> not x
True
>>> x.greenlets = [1]
>>> bool(x)
True
```

Is there a use-case where things like `if group` and `if not group` don't work?

PR #510, opened 2014-11-10
Inherit from Exception instead of BaseException

dbruns:
Base Exception is not meant to be inherited directly. "The base class for all built-in exceptions. It is not meant to be directly inherited by user-defined classes (for that, use Exception)."

### DIFF
--- a/gevent/timeout.py
+++ b/gevent/timeout.py
@@ -19,7 +19,7 @@ __all__ = ['Timeout',
            'with_timeout']
 
 
-class Timeout(BaseException):
+class Timeout(Exception):
     """Raise *exception* in the current greenlet after given time period::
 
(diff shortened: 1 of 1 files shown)


ekimekim:
The distinction here is important. Commonly error handling code will catch Exceptions, but not things such as KeyboardInterrupt, StopIteration or SystemExit, which all inherit directly from BaseException.

The question is whether a gevent.Timeout should be treated like an error, or allowed to raise to whichever caller is specifically looking for it.

While personally I like the current behaviour where error handling code can't "accidentally" catch my Timeout that I've made a specific check for, there's a far stronger argument: Backwards compatibility. Changing this would likely break alot of code in ways that wouldn't surface until things started timing out, at which point all hell would break loose.

PR #511, opened 2014-11-11
Pass joinall count argument to inner wait call

leth:


jamadden:
Closed with commit a755c446. Thank you.

PR #512, opened 2014-11-19
count is not working when raise_error is False

adieu:
Should pass `count` param to `wait` function when `raise_error` is False.

I have spent several days to debug a wired behavior of my program caused by this bug. Hope this could save other people some time in similar situation.

### DIFF
--- a/gevent/greenlet.py
+++ b/gevent/greenlet.py
@@ -410,7 +410,7 @@ def _kill(greenlet, exception, waiter):
 
 def joinall(greenlets, timeout=None, raise_error=False, count=None):
     if not raise_error:
-        wait(greenlets, timeout=timeout)
+        wait(greenlets, timeout=timeout, count=count)
     else:
         for obj in iwait(greenlets, timeout=timeout):
             if getattr(obj, 'exception', None) is not None:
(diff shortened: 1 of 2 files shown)


thedrow:
Can you please explain why this is required?
Also can someone review this?

jamadden:
Good catch, thank you! The only thing missing is a test case. Can you add one please? I would suggest something like this in `test__greenlet.py` in `TestStuff` aside the other `joinall` tests (I've verified this fix with this test case locally, but it would be nice to have it part of the commit, and give it a chance to run on Travis):

```python
    def test_joinall_count_raise_error(self):
        # When joinall is asked not to raise an error, the 'count' param still
        # works.
        def raises_but_ignored():
            raise ExpectedError("count")
        def sleep_forever():
            while True:
                sleep(0.1)

        sleeper = gevent.spawn(sleep_forever)
        raiser = gevent.spawn(raises_but_ignored)

        gevent.joinall([sleeper, raiser], raise_error=False, count=1)
        assert_ready(raiser)
        assert_not_ready(sleeper)

        # Clean up our mess
        sleeper.kill()
        assert_ready(sleeper)
```

adieu:
Sure, I'll add it to the test.

adieu:
Looks like `apt-get -qq -y update` is failing in travis. How can I trigger a rebuild for this PR?

jamadden:
I re-triggered the build and it still failed for the same reasons. But at least PyPy passed on Travis, and it passes locally, so that's good enough for me.

jamadden:
Thank you!

adieu:
Nice! :smile:

PR #528, opened 2015-01-06
Add full traceback to joinall exception raise

3ddi:
All explanation in rodolfo3 commit.


3ddi:
The tests fail on routines:SSL3_READ_BYTES:sslv3 alert certificate expired. Is the test server OK?

PR #533, opened 2015-01-20
Bugfix: Python3 SSLSocket missing imports.

teemuy:
SSLSocket was broken on Python3 due to missing imports.

```python
  File "/usr/local/lib/python3.3/http/client.py", line 1206, in connect
    server_hostname=server_hostname)
  File "gevent/_ssl3.py", line 50, in wrap_socket
    _context=self)
  File "gevent/_ssl3.py", line 91, in __init__
    if sock.getsockopt(SOL_SOCKET, SO_TYPE) != SOCK_STREAM:
NameError: global name 'SOL_SOCKET' is not defined
```

### DIFF
--- a/gevent/_ssl3.py
+++ b/gevent/_ssl3.py
@@ -13,7 +13,7 @@ import ssl as __ssl__
 _ssl = __ssl__._ssl
 
 import errno
-from gevent.socket import socket, timeout_default
+from gevent.socket import socket, timeout_default, SOL_SOCKET, SO_TYPE, SOCK_STREAM
 from gevent.socket import error as socket_error
 
 


thedrow:
This change sounds reasonable but how come it wasn't failing the build?
+1 on merging this anyway.
@teemuy Can you cover this with a test or should I open an issue about it?

jamadden:
In the past, the SSL tests were not running due to an earlier import error in the test code itself. I fixed that it 3db30e9, and the SSL tests pass under Python 3.3.

These constants are dynamically imported [at the top of the file](https://github.com/gevent/gevent/blob/3db30e936062eb6a8be86bd3d0c72bca39069da5/gevent/_ssl3.py#L26) and don't need to be manually imported.

PR #542, opened 2015-03-18
subprocess.Popen `bufsize` option actually used

romuald:
The `bufsize` option of the `subprocess.Popen` module wasn't used (at all), meaning that the default was usually buffered (platform dependant), and was also not modifiable.

This was also different from the behavior of the vanilla `subprocess.Popen` module

Note that this change the default behavior of the class (buffered becomes unbuffered in most cases)


jamadden:
LGTM. The change in buffering behaviour is a compatibility concern, but it looks like a small one to me, especially since this gets us closer to matching the standard library. Thank you!

PR #544, opened 2015-03-19
Make DummySemaphore.acquire return True

mouadino:
Usually gevent.lock.DummySemaphore and gevent.lock.Semaphore are used interchangeably
and since gevent.lock.Semaphore.acquire return a boolean for checking if acquiring
lock failed or worked, it will make sense to just return True for DummySemaphore.acquire,
this way we can just use afterward ``if not acquired: ...``.


thedrow:
+1

Issue #477, opened 2014-09-12
ssl broken for python > 2.7.9


joequant:
https://github.com/eventlet/eventlet/issues/135

2.7.9 removed _sslwrap which means that the gevent ssl.py will fail.

josegonzalez:
I had this same issue, but was able to correct it by adding @Eugeny's fix as a custom `ssl_279.py` module in my app and then importing it. Seems to fix my flask/gevent/gunicorn app at least.

josegonzalez:
As @pigmej mentioned, the SSLSocket issue is still a problem. Here's the "bad" [code](https://github.com/gevent/gevent/blob/master/gevent/_ssl2.py#L58). I looked into the gevent code, don't know what the exact fix would be. Lame that this effects heroku apps :(

mtsgrd:
@josegonzalez this hasn't affected our Heroku apps in my domain yet. Can you confirm you see it on Heroku?

As of yesterday, I now see this issue on Travis.

josegonzalez:
I definitely saw this on heroku when I commented here last, so I downgraded my python to 2.7.8 to have it work.

iElectric:
Debian has a patch: https://bazaar.launchpad.net/~ubuntu-branches/ubuntu/utopic/python-gevent/utopic/view/head:/debian/patches/sslwrap.py

tomster:
@iElectric you sure that patch works? i applied it to gevent 1.0.1 and that results in ``NameError: global name 'SSLContext' is not defined`` when running the patched version. looking at the initial file i cannot see how that patch should fix the issue.

does anybody have a fix for this?

iElectric:
Could you paste the whole traceback? Thanks!

fdr:
This is breaking https://github.com/wal-e/wal-e tests, after I upgraded to Ubuntu 14.10 from 14.04. I haven't seen this reported in the wild from a user, though.

iElectric:
@tomster https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/gevent_sslwrap.patch (also see above commit)

iElectric:
@fdr wal-e is still failing due to boto: https://gist.github.com/ocharles/b1b53355b5bb2162871e

itaykahana:
@pigmej I have the same issue with Heroku/Python 2.7.9/Gevent environment. Tried your patch, and also to change the SSLSocket constructor to accept `server_hostname` and `_context`.

[...]

Issue #513, opened 2014-11-20
ssl broken on Debian (and derivatives), as of python 2.7.8-12


csillag:
In v2.7.8-12 of the Debian python suit, which was released 3 days ago, they added [a patch](https://bugs.debian.org/cgi-bin/bugreport.cgi?msg=17;filename=python2.7-nossl3.patch;att=1;bug=768611), which, according to the [changelog](http://metadata.ftp-master.debian.org/changelogs//main/p/python2.7/python2.7_2.7.8-12_changelog),  "Allow building and testing without SSLv3 support"

In fact it removes many of the SSLv3-related constants, including `ROTOCOL_SSLv3`, which make gevent fail with a message similar to this:

```
  File "<whatever>/h/local/lib/python2.7/site-packages/gevent/ssl.py", line 386, in <module>
    def get_server_certificate(addr, ssl_version=PROTOCOL_SSLv3, ca_certs=None):
NameError: name 'PROTOCOL_SSLv3' is not defined
```

One could argue that it's not nice for a distribution to haphazardly remove constants that are still there in the upstream version, but these things happen, so... maybe gevent could do something about this?
(Like, maybe, move to  'PROTOCOL_SSLv23' instead?)

Starefossen:
Ok, thanks a lot for fixing this issue, but we need a new version with this fix in. Any ETA on v1.0.2 @denik?

csillag:
> Any ETA on v1.0.2 @denik or @Ivoz? 

+1

bastianl:
> Any ETA on v1.0.2 @denik or @Ivoz?

+1

pcdinh:
@denik It is time for a new release. Lots of people are waiting for it

AxelVoitier:
Yes, please!

This bug is biting anyone using gevent in a docker container using the official image for python, as it uses a debian image as a base...

aleksandr-vin:
I'm too on the waiting list, guys!

dbrgn:
+1 here

pedrudehuere:
+1 here

lmanolov:
+1

telbizov:
+1 here as well.

Jessie is now the Debian stable release as of a week ago. Just upgraded to it and my gevent based server experiences this problem.

Is there an ETA on releasing this in pip ?

baljanak:
+1

jonblack:
[...]

PR #546, opened 2015-03-21
Port Python 2.7.9 ssl.py

sprin:
Resolves #477.

In Python 2.7.9, we have new SSL interfaces, that are not quite
the old Python 2 interfaces, nor the same as the Python 3 interfaces.

This creates a new SSL module, _sslgte279.py, for Python 2 versions
greater than or equal 2.7.9, ported from Python 2.7.9's ssl.py:
https://hg.python.org/cpython/file/648dcafa7e5f/Lib/ssl.py.

The process of porting was done by starting with gevent's ssl3.py and
backporting Python 2.7.9's ssl.py.

Fixes these tests under Python 2.7.9:

 - /usr/local/bin/python -u -m monkey_test --Event test_urllib2net.py
 - /usr/local/bin/python -u -m monkey_test test_urllib2net.py
 - /usr/local/bin/python -u test__ssl.py

No new test failures. Still failing since 72119c8:

 - /usr/local/bin/python -u -m monkey_test --Event test_ssl.py
 - /usr/local/bin/python -u -m monkey_test test_ssl.py
 - /usr/local/bin/python -u test__pywsgi.py
 - /usr/local/bin/python -u test__makefile_ref.py
 - /usr/local/bin/python -u test__socket.py
 - /usr/local/bin/python -u test___example_servers.py
 - /usr/local/bin/python -u -m monkey_test --Event test_ftplib.py
 - /usr/local/bin/python -u test__socket_ssl.py
 - /usr/local/bin/python -u -m monkey_test test_ftplib.py
 - /usr/local/bin/python -u test__all__.py

In addition, basic HTTPS requests were tested with urllib3==1.10.2
and requests==2.6.0.

Issue #545, opened 2015-03-21
SSL_v3 removed in python 2.7.8-4


laurentL:
hi,
impossible to use gevent in debian jessie:

Traceback (most recent call last):
  File "/var/lib/jenkins/jobs/knockdaemonMaster/workspace/knockdaemon_test/TestAllSqlStuff.py", line 53, in setUp
    SolBase.voodoo_init()
  File "/var/lib/jenkins/jobs/knockdaemonMaster/workspace/.env/local/lib/python2.7/site-packages/pythonsol/SolBase.py", line 313, in voodoo_init
    monkey.patch_all(aggressive=aggressive)
  File "/var/lib/jenkins/jobs/knockdaemonMaster/workspace/.env/local/lib/python2.7/site-packages/gevent/monkey.py", line 190, in patch_all
    patch_ssl()
  File "/var/lib/jenkins/jobs/knockdaemonMaster/workspace/.env/local/lib/python2.7/site-packages/gevent/monkey.py", line 146, in patch_ssl
    patch_module('ssl')
  File "/var/lib/jenkins/jobs/knockdaemonMaster/workspace/.env/local/lib/python2.7/site-packages/gevent/monkey.py", line 62, in patch_module
    gevent_module = getattr(__import__('gevent.' + name), name)
  File "/var/lib/jenkins/jobs/knockdaemonMaster/workspace/.env/local/lib/python2.7/site-packages/gevent/ssl.py", line 386, in <module>
    def get_server_certificate(addr, ssl_version=PROTOCOL_SSLv3, ca_certs=None):
NameError: name 'PROTOCOL_SSLv3' is not defined

krak3n:
@laurentL see #513

thedrow:
@laurentL Yup. This issue should be closed as it's a duplicate of the already resolved #513.

jamadden:
Closing as duplicate of #513.

PR #551, opened 2015-04-16
Per gevent/gevent#546, rebase the changes on top of gevent 1.0.2. Fixes gevent/gevent#477.

jamadden:
I tested on both 2.7.8 and 2.7.9 with these results:

```
2/120 unexpected passes
 - /Python -u -m monkey_test --Event test_ssl.py
 - /Python -u -m monkey_test test_ssl.py

6/120 tests failed in 01:18

6/120 unexpected failures
 - /Python -u -m monkey_test --Event test_httpservers.py
 - /Python -u test_ares_timeout.py
 - /Python -u test__socket_dns6.py
 - /Python -u test__socket_dns.py
 - /Python -u -m monkey_test test_httpservers.py
 - /Python -u test__all__.py
```

The `httpservers` failure is because it relies on the internals of the CGI module that have now been moved. The DNS failures are due to differences in DNS configuration. `test_ares_timeout` is because it tries to bind to port 53, which doesn't work for non-root users under unix, and to detect this it relies on a specific error message which has either changed or varies between platforms. `test__all__` complains about `test_libev`, and I have no idea what that's about.

Notably, the `test_ssl` passes now on both 2.7.8 and 2.7.9. I didn't mess with the known_failures list because I didn't test under 2.6.


sprin:
:clap:

I am happy to review this weekend.

It sounds like `test_ares_timeout` should just be skipped by default if it tries to bind to a privileged port. It might make sense to keep it enabled if the tests were always running in a defined container/VM, but having a test that requires root does not make sense when most devs are probably running tests in an unprivileged venv.

We should look at the other tests to see if it makes sense to fix or disable them. I would suggest that tests that are sensitive to DNS config also don't make sense in the typical case.

jamadden:
Disabling `test_ares_timeout` is what I eventually did in #552...although, looking more closely at the underlying ares code now, it seems like there might be a way to actually specify a port to use, so it may be possible to make it pass. I'll look into that.

I keep saying the DNS tests are config/remote server dependent because sometimes I see them pass and sometimes (usually) they don't. I'd hate to disable them entirely, though, because it seems like they're testing an important part of the codebase (they run, just get slightly different output than expected). Perhaps there's a way to make them more tolerant of differences but still check the codepaths. I'll look into that too.

If I find anything I'll open a different PR.

jamadden:
Oh, also, I'm pretty sure the `test__all__` failure with 'libev' is just an artifact of re-using a build directory for master on PyPy and this branch on CPython; a pseudo-module gets created for the benefit of PyPy on master that I hadn't cleaned up.

jamadden:
I did find a better way to fix `test_ares_timeout`, `test_httpservers` and the DNS tests; I put those in a PR against master to keep this one as small as possible. But if requested I can put those changes here too.

In #477 some comments state that certain distributions have backported the SSL changes *without* changing the python version number (i.e., it's still 2.7.8). It would probably be easy to make this `ssl` module detect that and not just be driven off the version number. I can add that if desired, but it might be tricky to verify.

Please let me know if I can help move this forward in any way. In addition to #477, this should resolve #513 and #545.

jamadden:
@sprin and @ivoz any thoughts?

jamadden:
@sprin  @Ivoz Is this the major blocker for a 1.0.2 release? If so, we should probably try to get this out there sooner rather than later. (I'm not aware of a roadmap, that's something that should probably be discussed at some point.)

jamadden:
@sprin @ivoz I skipped the style checks to run the actual tests on Travis. Python 2.6 is clean. Python 2.7 had one timing related failure that is *probably* transient. Python 2.5 is currently blocked on python-greenlet/greenlet#82 and won't run; our options, as I see them, are to wait for a new release of greenlet, or to pin it to 0.4.5 under Python 2.5 (not good because it clashes when it does get fixed) or ignore the issue.

jamadden:
The build is now green across the board for 2.5, 2.6 and 2.7. Unless there are any objections, I will probably merge this soon.

thedrow:
Woot! Well done.

guilbep:
Yeahh gevent 1.0.2 fixed it :) :+1:

PR #558, opened 2015-05-18
SSL fixes for python 2.7.9

### LINKED ISSUE #477: ssl broken for python > 2.7.9

ml31415:
This fixes #477 based  on https://github.com/Eugeny/gevent/commit/2c5302bed48e96db3d68169d6ae9eab36f5aa330 and additionally also fixes the socket connect method, which seemd to be left out there. I'd suggest to either take this request or adjust the other patch accordingly, before merging this into the new release.

### DIFF
--- a/gevent/_ssl2.py
+++ b/gevent/_ssl2.py
@@ -80,15 +80,28 @@ class SSLSocket(socket):
             self._sslobj = None
         else:
             # yes, create the SSL object
-            if ciphers is None:
-                self._sslobj = _ssl.sslwrap(self._sock, server_side,
-                                            keyfile, certfile,
-                                            cert_reqs, ssl_version, ca_certs)
+            if hasattr(_ssl, 'sslwrap'):
+                if ciphers is None:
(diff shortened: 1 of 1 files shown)


jamadden:
@ml31415 The code currently merged into the 1.0.2 and master branches went in a different direction than https://github.com/Eugeny/gevent/commit/2c5302bed48e96db3d68169d6ae9eab36f5aa330. Essentially, The `ssl` module from 2.7.9 was backported into `_sslgte279.py` and is used only under 2.7.9+.  `_ssl2` is renamed from `ssl.py` and is otherwise unchanged; it is used on <= 2.7.8 (including 2.6 and 2.5). This increases compatibility with the running version of Python by providing the full `ssl` interface of the natively running version.

ml31415:
I'm totally fine with closing this pull request in this case. I just wanted to make sure, no erroneous calls to sslwrap remain in the code.

Issue #461, opened 2014-07-25
Can't catch KeyboardInterrupt error on python3


jander:
When starting a stream server on python3. gevent can't catch KeyboardInterrupt error to exit.
for example:
```
def start_server(address, handle_func):
    server = StreamServer(address, handle_func)
    print('Starting %s on %r' % ( __file__, address))
    try:
        server.serve_forever()
    except (KeyboardInterrupt, SystemExit):
        print('Exiting...')
    finally:
        server.stop()
```
`print('Exiting...')` never run on python3.

I look over `hup.py`, find line 263 is:
```
NOT_ERROR = (GreenletExit, SystemExit)
```
I think it should be:
```
NOT_ERROR = (GreenletExit, SystemExit, KeyboardInterrupt)
```

I changed that line code on my pc, and the server run correctly.

hashstat:
#471 appears to be a duplicate of this issue and seems to have more activity.

jamadden:
Fixed by #578.

PR #564, opened 2015-06-01
Fix issues #461 and #471: Exception raised in print_exception().

hashstat:
If the exception value is None, the exception type name is printed to stderr (like Python 2.x); otherwise, pass it off to traceback.print_exception(). This allows KeyboardInterrupt to properly propagate up the stack.


jamadden:
Can you rebase on current master (Travis CI tests should be more reliable), and add a descriptive comment and a test case please?

PR #572, opened 2015-06-12
Fix invalid syntax in _socket3.py

parasyte:
$ pyflakes gevent/_socket3.py
    gevent/_socket3.py:135: invalid syntax
        def makefile(self, mode="r", buffering=None, *,
                                                       ^

### DIFF
--- a/gevent/_socket3.py
+++ b/gevent/_socket3.py
@@ -132,7 +132,7 @@ class socket(_socket.socket):
             sock.setblocking(True)
         return sock, addr
 
-    def makefile(self, mode="r", buffering=None, *,
+    def makefile(self, mode="r", buffering=None,
                  encoding=None, errors=None, newline=None):
         """makefile(...) -> an I/O stream connected to the socket
 


jamadden:
I believe that is the correct function declaration. From [the documentation](https://docs.python.org/3/library/socket.html#socket.socket.makefile):
```
socket.makefile(mode='r', buffering=None, *, encoding=None, errors=None, newline=None)
      Return a file object associated with the socket....
```
While that's not legal syntax in Python2, it is in Python3, where it serves to mark the beginning of [keyword-only arguments](https://docs.python.org/3/reference/compound_stmts.html#function-definitions).

gevent has a customized way to invoke pyflakes to account for differences between versions. The way to run pyflakes is like this:
```
$ python util/pyflakes.py
```

I have a branch that runs the pyflakes and pep8 tests on Travis; it's [currently passing](https://travis-ci.org/gevent/gevent/jobs/65292684). I hope to get it merged soon.

parasyte:
Ok. If it makes any difference, I noticed the syntax error message when installing gevent with Python 2. It appears to be benign, but I didn't investigate why the file was being parsed by python2 during installation.

Maybe there's a better "fix" there.

PR #575, opened 2015-06-16
Use the TravisCI container infrastructure and enable linters

jamadden:
This comes with its own implicit speedups, but also allows use of caching the cython installation. Startup time goes from up to 30minutes to typically a few seconds.

This lets us use the travis test matrix more effectively. We spin off a separate process early on to run all the linting checks for code quality and garbage cycles. These should all be passing now and be maintained in a passing fashion. The linters were added to tox.ini for local developer convenience.

### DIFF
--- a/.pep8
+++ b/.pep8
@@ -1,4 +1,4 @@
 [pep8]
-ignore=E702,E265,E402,E731,E266
+ignore=E702,E265,E402,E731,E266,E261
 max_line_length=160
-exclude=.git,build,2.6,2.7,2.7pypy,3.3,test_support.py,test_queue.py,patched_tests_setup.py,test_threading_2.py,lock_tests.py,_sslgte279.py
+exclude=.tox,.git,build,2.6,2.7,2.7pypy,3.3,test_support.py,test_queue.py,patched_tests_setup.py,test_threading_2.py,lock_tests.py,_sslgte279.py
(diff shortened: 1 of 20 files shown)


jgehrcke:
Great work, Jason. I am curious: what is the current organizational state -- can you considered to be the new maintainer of gevent? If that is the case, I'd love to get in touch. How much can Denis still contribute?

jamadden:
I wouldn't consider myself "the maintainer". I and a few others have commit access to the repository, and I have access to deploy to PyPI. But there are things I don't have access to do, like update the documentation or website. All of us are just volunteers working on things as-and-when, and that includes Denis (though I don't want to put any words into his mouth).

jgehrcke:
Thanks for your input! I got the impression that Denis was, until quite recently, the only person with commit access, which also was one of the reasons why the project development slowed down a bit. I am under the impression that things changed recently (positive!), but that this was not communicated transparently. Really, no offense, this is just my impression and I hope you can provide input on that :-). I first contributed to gevent in 2012, followed its development quite closely since then, and am quite willing to contribute again. So, I am really interested in understanding who "officially" maintains this project at the moment. Getting to know the people is important. Even if at the moment that just means to know who has commit access (an information which GitHub "hides"). We should probably move this to the mailing list ... Cheers!

PR #576, opened 2015-06-17
Py3 compatibility

smurfix:
With this (and related changes in gevent-socketio and -websocket),
at least the simple socketio chat example works with py3.

### DIFF
--- a/gevent/_socket3.py
+++ b/gevent/_socket3.py
@@ -3,6 +3,7 @@
 import io
 import time
 from gevent import _socketcommon
+from gevent.hub import text_type
 import _socket
 from io import BlockingIOError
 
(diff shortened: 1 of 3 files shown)


gevent/_socket3.py
```diff
@@ -22,6 +23,8 @@
 
 
 def _get_memory(string, offset):
+    if isinstance(string,text_type):
+        string = string.encode('utf-8')
```
jamadden: Python 3 doesn't support sending (unicode) strings across a socket, and I don't think gevent should introduce a gratuitous difference here. Instead, we should update the examples to add the appropriate `b` prefixes.

```
Python 3.4.3 (default, May 25 2015, 18:48:21)
[GCC 4.2.1 Compatible Apple LLVM 6.0 (clang-600.0.56)] on darwin
>>> import socket
>>> s = socket.socket()
>>> s.connect(('www.yahoo.com', 80))
>>> s.sendall('GET / HTTP/1.0\r\n\r\n')
TypeError: 'str' does not support the buffer interface
```

jamadden:
#412 and related pulls were merged with #581, so pywsgi should be working now. Thank you for looking into this! Please open a new issue/pr if you still have problems.

PR #578, opened 2015-06-18
Fix for issues #461 and #471.

hashstat:
This handles None exception values that would otherwise fail when passed
to traceback.print_exception() and prevent the process from exiting. An associated test is also included.

Related to #564.

### DIFF
--- a/greentest/test__issues461_471.py
+++ b/greentest/test__issues461_471.py
@@ -10,6 +10,7 @@ import sys
 
 if sys.argv[1:] == ['subprocess']:
     import gevent
+
     def task():
         sys.stdout.write('ready\n')
         sys.stdout.flush()
@@ -20,7 +21,9 @@ if sys.argv[1:] == ['subprocess']:
         pass
 else:
     import signal
-    from subprocess import Popen, PIPE, TimeoutExpired
+    from subprocess import Popen, PIPE
+    import time
+
     if sys.platform.startswith('win'):
         from subprocess import CREATE_NEW_PROCESS_GROUP
         kwargs = {'creationflags': CREATE_NEW_PROCESS_GROUP}
@@ -29,9 +32,11 @@ else:
     p = Popen([sys.executable, __file__, 'subprocess'], stdout=PIPE, **kwargs)
     p.stdout.readline()
     p.send_signal(signal.SIGINT)
-    try:
-        p.wait(3)
-    except TimeoutExpired:
+    for i in range(30):
+        if p.poll() is not None:
+            break
+        time.sleep(0.1)
+    else:
         p.terminate()
         sys.exit(1)
     sys.exit(p.returncode)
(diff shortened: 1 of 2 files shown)


jamadden:
Wow, nice integration test! 

Unfortunately, it uses an exception only available on Python >= 3.3, so all the Python 2 builds failed. Can you address this please?

```
| GEVENT_RESOLVER=thread /home/travis/virtualenv/python2.7.9/bin/python -u test__issues461_471.py
    File "test__issues461_471.py", line 23, in <module>
      from subprocess import Popen, PIPE, TimeoutExpired
  ImportError: cannot import name TimeoutExpired
```

greentest/test__issues461_471.py
```diff
@@ -0,0 +1,37 @@
...
+        kwargs = {}
+    p = Popen([sys.executable, __file__, 'subprocess'], stdout=PIPE, **kwargs)
+    p.stdout.readline()
+    p.send_signal(signal.SIGINT)
+    try:
+        p.wait(3)
+    except TimeoutExpired:
+        p.terminate()
+        sys.exit(1)
+    sys.exit(p.returncode)
```
jgehrcke: Just a couple of idea.s

For making this Py2/3-compatible, depending on when in the test suite this test is about to run (after Timeout tests it should be!), it might be a good idea to just use `gevent.Timeout` with a polling-based approach as a *simple* cross-platform timeout solution, like so (untested):

```python
rc = None        
with gevent.Timeout(3, False):
    while rc is None:
        gevent.sleep(0.01)
        rc = p.poll()
    sys.exit(rc)
p.terminate()     
sys.exit(1)
```

You might also want to do `assert p.stdout.readline() == "ready"`, when you're readline()ing anyway. I am not 100 % sure right now, but if there is a chance for a race condition among poll() and the Timeout exception, the argument to the last `sys.exit()` should not simply be 1, but sth like `rc if rc is not None else 1`. However, I think that should never be of relevance, considering the large timeout value of three seconds.

hashstat:
Sorry about that. I should have tested with 2.7 before.

greentest/test__issues461_471.py
```diff
@@ -0,0 +1,42 @@
...
+        kwargs = {}
+    p = Popen([sys.executable, __file__, 'subprocess'], stdout=PIPE, **kwargs)
+    p.stdout.readline()
+    p.send_signal(signal.SIGINT)
+    for i in range(30):
+        if p.poll() is not None:
+            break
+        time.sleep(0.1)
+    else:
+        p.terminate()
```
jgehrcke: On POSIX-compliant platforms, `terminate()` just sends a signal. We should still `wait()` after after that to make sure the child has really exited, It's best practice, like `join()` is in case of greenlets, threads, and `multiprocessing.Process` objects.

jamadden:
LGTM. Thank you.

PR #582, opened 2015-06-21
Decorate gevent.select.select with @staticmethod

smurfix:
The original select.select has that decorator, too, so it works as a method
(assigned to selectors.SelectSelector._select).

Discovered while porting flask.socketio to Python3.

### DIFF
--- a/gevent/select.py
+++ b/gevent/select.py
@@ -39,7 +39,7 @@ class SelectResult(object):
         self.write.append(socket)
         self.event.set()
 
-
+@staticmethod
 def select(rlist, wlist, xlist, timeout=None):
     """An implementation of :meth:`select.select` that blocks only the current greenlet.
 


jamadden:
Unfortunately, `staticmethod` objects are not callable. This change makes every Travis build fail with errors like the following:
```
    File "/opt/python/3.4.2/lib/python3.4/socketserver.py", line 286, in handle_request
      fd_sets = _eintr_retry(select.select, [self], [], [], timeout)
    File "/opt/python/3.4.2/lib/python3.4/socketserver.py", line 154, in _eintr_retry
      return func(*args)
  TypeError: 'staticmethod' object is not callable
```

Although a more general solution would be better, perhaps [this change in master](https://github.com/gevent/gevent/commit/cd483e90e627b583839690deb4c744fbd48a9d19#diff-ad8b9019feec8da668ef2771a5661436R205) fixes the issue you were seeing?

smurfix:
On 22.06.2015 09:13, Jason Madden wrote:
> ||
>
> Although a more general solution would be better, perhaps this change
> in master
> <https://github.com/gevent/gevent/commit/cd483e90e627b583839690deb4c744fbd48a9d19#diff-ad8b9019feec8da668ef2771a5661436R205>
> fixes the issue you were seeing?
>
Yes, that would fix it. Thanks.

-- 
-- Matthias Urlichs

jamadden:
Ok, Closing for now then.

PR #589, opened 2015-06-25
Add original tracebacks when re-raising an exception from another greenlet

jamadden:
This is an updated version of #528 and #450 that passes all the tests.

While it generally does seem like a very useful capability, I'm slightly concerned that obscuring the `get` or `wait` part of the traceback might in itself be confusing.

I'm also a little concerned about the potential for reference cycles. A number of the tests had to be modified to avoid them. (Simply always deleting the traceback when the exception is re-raised isn't sufficient because there are places that the exception doesn't actually get raised, I think.)

Still, I'm inclined to merge this as an "experimental" feature to get it in the hands of users of an alpha/beta/rc release and see how the experience goes, unless there is more discussion that takes place here.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -24,6 +24,10 @@ Unreleased
 - ``gevent.queue.JoinableQueue`` treats ``items`` passed to
   ``__init__`` as unfinished tasks, the same as if they were ``put``.
   Initial PR #554 by DuLLSoN.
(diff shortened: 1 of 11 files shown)


jamadden:
The reference issue was solved in a cleaner way with a subsequent commit that pickles the traceback.

Issue #602, opened 2015-07-02
Support patching select.poll


3ddi:
Being used in many python libraries, and is supported by eventlet.

jamadden:
Hmm, in eventlet's [select module](https://github.com/eventlet/eventlet/blob/master/eventlet/green/select.py) I don't see any evidence that it supports `epoll`. Am I looking in the wrong place?

Pull requests welcome!

3ddi:
You are right, they simply don't use aggressive removal of poll/epoll by default. so you get a patched select.poll witch is not green.
Nevertheless, I'm adding a green poll support.

jamadden:
Merged in 722f833

PR #604, opened 2015-07-08
Added support for a green poll (includes monkey patching)

3ddi:
Fix for issue #602

### DIFF
--- a/gevent/select.py
+++ b/gevent/select.py
@@ -3,11 +3,13 @@ from __future__ import absolute_import
 from gevent.event import Event
 from gevent.hub import get_hub
 from gevent.hub import integer_types
+from select import POLLIN, POLLOUT
 
-__implements__ = ['select']
+__implements__ = ['select', 'poll']
 __all__ = ['error'] + __implements__
 
 import select as __select__
+
 error = __select__.error
 
 
@@ -16,14 +18,13 @@ def get_fileno(obj):
         fileno_f = obj.fileno
     except AttributeError:
         if not isinstance(obj, integer_types):
-            raise TypeError('argument must be an int, or have a fileno() method: %r' % (obj, ))
+            raise TypeError('argument must be an int, or have a fileno() method: %r' % (obj,))
         return obj
     else:
         return fileno_f()
 
 
 class SelectResult(object):
-
     __slots__ = ['read', 'write', 'event']
 
     def __init__(self):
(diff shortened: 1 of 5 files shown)


jamadden:
Can you accomplish this without modifying `greentest/3.4/test_selectors.py`? That's a direct copy of the CPython source code; for full compatibility with CPython non-monkey-patched systems we don't modify the files in the 2.x/3.x directories if at all possible.

3ddi:
Reverted the test as you asked.

jamadden:
@denik If you have some time, can you please share any design decisions you had made around the select module and only patching the `select` function? It seems like many of the stdlib functions will prefer `poll` if it's available (based on the earlier test failures). Do you have any comments or concerns on this implementation approach? Thanks!

thedrow:
This needs more tests with more usecases.
I don't think it's enough to add only one test to ensure that this feature works.
Other than that, well done. I looked into it myself and had no idea how to implement poll() in gevent.

3ddi:
It actually runs much more tests then one, many tests were skipped before (You can take a look at the failing tests). From what I've seen PollSelectorTestCase (about 10 tests), test_telnetlib (apparently telnet is using poll when it exists) and some more.

thedrow:
Ok, LGTM.

gevent/select.py
```diff
@@ -3,11 +3,18 @@
...
 from gevent.hub import integer_types
+from select import POLLIN, POLLOUT
 
-__implements__ = ['select']
+try:
+    from select import poll as original_poll
+except ImportError:
+    original_poll = None
+
+__implements__ = ['select', 'poll']
```
jamadden: If the original `select` module didn't implement `poll`, we probably shouldn't either.

3ddi:
1. Gevent poll can be used on systems without poll, why should we disable it? We already disabled the automatic monkey patching.
2. We don't need to remove the poll when poll is not available, because it's already not there.

jamadden:
Because the performance characteristics and scalability are likely to be very different than what a user of `poll` is likely to expect. If `poll` isn't available on the system, then libev is probably going to be using the select backend. I don't really see the point of adding the overhead of *pretending* to implement poll.

3ddi:
Removed poll in case it is not available in the system. What I meant is that we don't need to remove poll conditionally when monkey patching because if it doesn't exist we don't it.

3ddi:
Not sure if the last error has anything to do with the patch changes.

jamadden:
Probably not, `test__issue6` occasionally has problems on Travis. I would mark it as FLAKY, but i've never reproduced the problems locally and I don't understand what causes them. I've kicked off a re-run, that usually fixes things.

jamadden:
I was just about to merge this when I thought of something. Suppose that lots of events are becoming available one after another. This means that a call to `poll.poll` will wind up getting events from `self.poll_obj`, the blocking OS call. If a greenlet is in a `poll` loop, and, when that greenlet handles an event, it doesn't yield to the loop (make another gevent call), then as long as events keep arriving, that greenlet is starving all other greenlets. With the `select` implementation, I think you're guaranteed that the calling greenlet yields back to the loop.

This is *probably* an unlikely scenario (you're polling because you want to do I/O, after all, and those *should* be green calls), but I could see it arising, especially in a non-monkey-patched system. The use of the original polling object was just intended to be an optimization, right? It [looks like](https://travis-ci.org/gevent/gevent/builds/70434677) all the tests are going to pass without it.

3ddi:
Well, I that's not entirely correct, I implemented this because of some fd test, which demonstrated that libev can't handle a big amount of arriving events (64 was the limit if I'm correct).
Regarding the starvation, I can argue that excuse for many other green function gevent supplies, socket.send for instance (and many more), the users simply have to be responsible and use green sleeps to avoid starvation.

PR #606, opened 2015-07-09
pyp3 compat fixes

mbarkhau:
This fixes two errors which came up while testing with `pypy3-2.4.0-linux64` and `gevent-1.1a3.dev0-py3.2-linux-x86_64`.

The fix to `gevent/server.py` is for this error

      File "/home/mbarkhau/workspace/pyffm03/venv/site-packages/gevent-1.1a3.dev0-py3.2-linux-x86_64.egg/gevent/baseserver.py", line 171, in _do_read
        args = self.do_read()
      File "/home/mbarkhau/workspace/pyffm03/venv/site-packages/gevent-1.1a3.dev0-py3.2-linux-x86_64.egg/gevent/server.py", line 101, in do_read
        fd, address = sock._accept()
    _socket.error: [Errno 11] Resource temporarily unavailable
    <WSGIServer at 0x3c309f8 fileno=5 address=127.0.0.1:8000> failed with error

The fix to `gevent/threading.py` is for this error

      File "/home/mbarkhau/workspace/pyffm03/venv/site-packages/gunicorn/arbiter.py", line 507, in spawn_worker
        worker.init_process()
      File "/tmp/pip-build-9zxz8v/gunicorn/gunicorn/workers/ggevent.py", line 185, in init_process
        self.patch()
      File "/tmp/pip-build-9zxz8v/gunicorn/gunicorn/workers/ggevent.py", line 65, in patch
        monkey.patch_all(subprocess=True)
      File "/home/mbarkhau/workspace/pyffm03/venv/site-packages/gevent/monkey.py", line 294, in patch_all
        patch_thread(Event=Event)
      File "/home/mbarkhau/workspace/pyffm03/venv/site-packages/gevent/monkey.py", line 157, in patch_thread
        patch_module('threading')
      File "/home/mbarkhau/workspace/pyffm03/venv/site-packages/gevent/monkey.py", line 94, in patch_module
        gevent_module = getattr(__import__('gevent.' + name), name)
      File "/home/mbarkhau/workspace/pyffm03/venv/site-packages/gevent/threading.py", line 44, in <module>
        k, v = __threading__._active.items()[0]
    TypeError: 'dict_items' object is not subscriptable

### DIFF
--- a/gevent/pywsgi.py
+++ b/gevent/pywsgi.py
@@ -453,7 +453,7 @@ class WSGIHandler(object):
             finally:
                 # Avoid dangling circular ref
                 exc_info = None
-        self.code = int(status.split(' ', 1)[0])
+        self.code = int(status.split(b' ', 1)[0])
         self.status = status
         self.response_headers = headers
 
--- a/gevent/pywsgi.py
+++ b/gevent/pywsgi.py
@@ -453,7 +453,7 @@ class WSGIHandler(object):
             finally:
                 # Avoid dangling circular ref
                 exc_info = None
-        self.code = int(status.split(b' ', 1)[0])
+        self.code = int(status.split(' ', 1)[0])
         self.status = status
         self.response_headers = headers
 
--- a/gevent/server.py
+++ b/gevent/server.py
@@ -103,10 +103,6 @@ class StreamServer(BaseServer):
                 if not sock.timeout:
                     return
                 raise
-            except _socket.error as err:
-                if err.args[0] == EWOULDBLOCK:
-                    return
-                raise
             sock = socket(sock.family, sock.type, sock.proto, fileno=fd)
             # XXX Python issue #7995?
             return sock, address
--- a/gevent/server.py
+++ b/gevent/server.py
@@ -103,6 +103,10 @@ class StreamServer(BaseServer):
                 if not sock.timeout:
                     return
                 raise
+            except _socket.error as err:
+                if err.args[0] == EWOULDBLOCK:
+                    return
+                raise
             sock = socket(sock.family, sock.type, sock.proto, fileno=fd)
             # XXX Python issue #7995?
             return sock, address
--- a/gevent/threading.py
+++ b/gevent/threading.py
@@ -41,7 +41,7 @@ if PYPY:
     # otherwise we get a new DummyThread, which cannot be joined.
     # Fixes tests in test_threading_2
     if _get_ident() not in __threading__._active and len(__threading__._active) == 1:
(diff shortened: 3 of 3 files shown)


jamadden:
Unfortunately, the change to pywsgi breaks CPython 3.3 and 3.4; PyPy3 is the odd one out here, possibly because it is (mostly) an implementation of Python 3.2. And Python 3.2 is not currently supported by the code base, due to the difficulties in supporting 2.x and 3.2 in the same codebase (true, PyPy3 makes that easier by supporting the 3.3 syntax, but there are still likely to be library issues.)

mbarkhau:
I think the issue with pywsgi.py was my mistake. Now that I pass all parameters to start_response as strings instead of bytes, it seems to work fine.

As for the other issues, I have a hello worldish server running now without any errors using these two patches.

mbarkhau:
Hmm, the build is still failing for python3.4, but I'm having a hard time parsing the error log.

jamadden:
The threading change LGTM, I think we could take that without a test case. 

I'm a little concerned about the server.py change, though. That seems like a case where PyPy3 is either doing something wrong, or implementing the 3.2 behaviour, not the 3.3 behaviour. The lines above the added exception handler are supposed to be catching all the blocking cases (and do, under 3.3 and 3.4). That's going to make the new handler look redundant, and since we can't automate a test for it on Travis, it's likely to get blown away or broken with no notice. I'm also somewhat averse to peppering the code base with redundancies and changes *just* for the current version of PyPy3 (because we can't support CPython 3.2 without a fairly large patch) when, supposedly, PyPy3's 3.3 support is coming "soon." Lastly, I'm not sure how much good this one change will do; without being able to run the entire test suite in an automated fashion on PyPy3, how much good is partial, undocumented support in one module?

I really appreciate your interest in using gevent with pypy3. I'd like to be able to do so myself. But right now it seems like the drawbacks may be too much. Of course, I could just be making a false slippery slope argument. I'm open to counter-arguments, especially compelling use cases!

jamadden:
It's not unusual for `test__timeout` to fail when Travis is heavily loaded. A number of gevent tests are pretty timing sensitive. Let's see if it passes in the re-run I just kicked off.

mbarkhau:
Ok then I'll revert the change to `server.py`. I'm only working on a demo using pypy3 anyway, so I'm fine with python 3.2 not being supported and it seems that the subset of gevent which I'm using is working fine for that and I'll just use that patch locally. Hopefully pypy3.3 will be out soon.

jamadden:
I hope so too! Thank you.

Issue #543, opened 2015-03-19
Segmentation Fault on killing job repeatedly


jonahkichwacoders:
I am getting a segmentation fault when running gevent on this reduced test case. I am using:
gevent==1.0.1
greenlet==0.4.5
Python 2.7.8
on an Arm Linux box. The test works fine on my x86_64 Linux.

```python
import gevent

def worker():
    while True:
        gevent.sleep()
job = gevent.spawn(worker)

gevent.spawn(job.kill)
gevent.spawn(job.kill)
job.join()

```

The segmentation fault occurs when job.kill completes for the first time.

I am continuing to investigate, but any insight would be gratefully received.

jonahkichwacoders:
I have worked around this problem by turning off optimizations in greenlet and gevent.

The segfault is happening on callbacks.c:173 in function gevent_call:
The line is:
    Py_DECREF(callback);
at this point callback == 0 and prior to the call to PyObject_Call on line 161 the local variable callback was != 0.

RudiSchloesser:
Hi Jonah,
looks like I'm facing the same for PPC. I'm using bitbake and build of gevent 1.0.1 with bb-file from openembedded fails (configure in libev expects --host leading to missing config.h). Setting up the environment and running configure manually solves this problem and build of gevent succeeds. But still SIGSEGV in hub.py. How to force cython to leave out -O2 in the CFLAGS for the GCC?

jonahkichwacoders:
I had difficulty figuring out how to do it "properly", so in the end I went the patch setup.py route. As I am using buildroot, I am using its patching mechanism http://buildroot.net/downloads/manual/manual.html#customize-patches 

I have two patches, one for greenlet 0.4.5:

```patch
diff --git a/setup.py b/setup.py
index 1844948..34e919a 100755
--- a/setup.py
+++ b/setup.py
@@ -11,6 +11,9 @@ if ((sys.platform == "openbsd4" and os.uname()[-1] == "i386")
     or ("SunOS" in platform.platform() and platform.machine() == "sun4v")):
     os.environ["CFLAGS"] = ("%s %s" % (os.environ.get("CFLAGS", ""), "-Os")).lstrip()
 
+# Turn off all optimizations
+os.environ["CFLAGS"] = ("%s %s" % (os.environ.get("CFLAGS", ""), "-O0")).lstrip()
+
 try:
     if not (sys.modules.get("setuptools")
             or "develop" in sys.argv

```

one for gevent 1.0.1:

[...]

PR #616, opened 2015-07-16
for 32-bit build on 64-bit linux, grab CFLAGS -m32 from python.sysconfig

lanstin:


jamadden:
Hmm, I'm not sure about this. Won't that override the value that the user has set in the environment? E.g., some ARM users are finding that they need to disable optimizations by setting -O0 (#543) (sometimes I need to do that when debugging).

lanstin:
I only look at whether or not -m32 was used in the python build, not any other compiler settings.  Checking the build failure.

lanstin:
Fixing build failures, so will push another commit to PR (or would you prefer new PR with squashed?)  It seems Python 2.6 lacks sysconfig module, so I'll need to think about that.  

Also, I can make it so that we preserve existing CFLAGS env variable, and just add -m32 in that case.

jamadden:
> or would you prefer new PR with squashed?

Personally, I'm not big on squashing away history. Yes it makes things "cleaner", but it also throws out the steps that it took to get from non-working to working, and sometimes those steps contain important lessons.

lanstin:
Everything is green but the testing build with python 3.4 which has this error:

                   stderr=STDOUT)
  Expected:
      b'ls: non_existent_file: No such file or directory\n'
  Got:
      b'ls: cannot access non_existent_file: No such file or directory\n'

Which I believe is a change to the /bin/ls program not python per se.  

https://github.com/goj/coreutils/commit/3a208b14048121755eb01377f985372561a4b953#diff-3a7e15506b2a0bc55177b8d159cf98a8R2614

So I think it's ok.  It will not remove items from the CFLAGS (and this change only affects config time for c-ares, not compile time nor any other config time).

PR #634, opened 2015-08-11
timeout is not what you think it is

mikekap:
It's the exception class. I believe the intended thing here is self.timeout.

### DIFF
--- a/gevent/_socket2.py
+++ b/gevent/_socket2.py
@@ -290,7 +290,7 @@ class socket(object):
         try:
             return sock.send(data, flags)
         except error as ex:
-            if ex.args[0] != EWOULDBLOCK or timeout == 0.0:
+            if ex.args[0] != EWOULDBLOCK or self.timeout == 0.0:
                 raise
             sys.exc_clear()
             self._wait(self._write_event)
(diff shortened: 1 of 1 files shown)


gevent/_socket2.py
```diff
@@ -290,7 +290,7 @@ def send(self, data, flags=0, timeout=timeout_default):
         try:
             return sock.send(data, flags)
         except error as ex:
-            if ex.args[0] != EWOULDBLOCK or timeout == 0.0:
+            if ex.args[0] != EWOULDBLOCK or self.timeout == 0.0:
```
jamadden: I don't believe this change is correct. `timeout` is a keyword argument to this function; if it's isn't set, it's already replaced with `self.timeout`:
```
    def send(self, data, flags=0, timeout=timeout_default):
        sock = self._sock
        if timeout is timeout_default:
            timeout = self.timeout
        try:
           ...
```

gevent/_socket2.py
```diff
@@ -381,7 +381,7 @@ def sendto(self, *args):
         try:
             return sock.sendto(*args)
         except error as ex:
-            if ex.args[0] != EWOULDBLOCK or timeout == 0.0:
+            if ex.args[0] != EWOULDBLOCK or self.timeout == 0.0:
```
jamadden: This does appear to be a legitimate typo. _socket3.py has the same issue in its `sendto` method.

PR #665, opened 2015-10-15
Include Python3.5 in AppVeyor CI Matrix

hexchain:


hexchain:
Oh, sorry for the trailing whitespace.

jamadden:
Unfortunately, it [doesn't even build](https://ci.appveyor.com/project/denik/gevent/build/1.0.147/job/9c59l0a88utvxc3q) on AppVeyor's 3.5, failing for different reasons under 32- and 64-bits. At least one of the errors looks like it's in the standard platform headers itself, so I don't know where the actual problem lies.

jamadden:
I would like to see appveyor builds on 3.5, because that's where we get the binary wheels that go on PyPI. Unfortunately, I have neither sufficient knowledge of building extensions on Windows nor the patience necessary to get it to debug these build issues. More help would be appreciated.

jamadden:
The https://github.com/ogrisel/python-appveyor-demo repo had part of the answer, and the rest of it looks like it's due to outdated code in libev's WIN32 support (although the error from a standard library header is misleading at best and sloppy on MS's part at worst). There are a few other things that need changed too. I'm working on a complete fix in another branch. Thanks for the PR!

jamadden:
For future Googler's: Here's the error with libev, when compiled with MSVC 14 (Microsoft Visual Studio 14 or 2014):
```
C:\Program Files (x86)\Windows Kits\10\include\10.0.10240.0\ucrt\sys/timeb.h(25): error C2032: '__timezone': function cannot be member of struct '__timeb32' 
C:\Program Files (x86)\Windows Kits\10\include\10.0.10240.0\ucrt\sys/timeb.h(33): error C2032: '__timezone': function cannot be member of struct '__timeb64' 
C:\Program Files (x86)\Windows Kits\10\include\10.0.10240.0\ucrt\sys/timeb.h(42): error C2032: '__timezone': function cannot be member of struct 'timeb' 
libev\ev.c(2329): warning C4133: '=': incompatible types - from 'WSABUF *' to 'CHAR *' 
```
The cause was that libev's `ev_win32` included a no-longer-necessary file: `#include <sys/timeb.h>`. Removing that include fixed the issue.

PR #672, opened 2015-10-25
use object.__getattribute in gevent.local instead.

Ivan-Zhu:
when an application define a subclass derived from gevent.local, if
directly use self.xxx get attribute, gevent can't guarantee the
__getatrribute__ behave.

usecase like the follow:
https:review.openstack.org/#c/1022/3/nova/local.py

use gevent instead of eventlet in the usecase, it will work under
gevent 1.0.1 but won't work under genvet 1.0.2. Although we can
work around this in app level, keeping compatibility in gevent is
better.

### DIFF
--- a/gevent/local.py
+++ b/gevent/local.py
@@ -217,7 +217,7 @@ def _patch(self):
         dct = impl.create_dict()
         args, kw = impl.localargs
         with impl.locallock:
-            self.__init__(*args, **kw)
+            object.__getattribute__(self, '__init__')(*args, **kw)
     with impl.locallock:
         object.__setattr__(self, '__dict__', dct)
         yield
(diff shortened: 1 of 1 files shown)


jamadden:
The link you provided doesn't work so I can't see the use case. 

This would need a test case that fails without the change but passes with the change.

Because the implementation of local comes directly from the standard library of 3.4, it seems likely that the test case would fail on the standard library too, though. That suggests that this is falling into the realm of implementation defined behaviour, in which case I would prefer to keep gevent's behaviour as close to the standard library as possible.

jamadden:
Thank you for opening this PR. Since there's been no forward motion on it for about 5 months, I'm going to go ahead and close it on the basis of the previous comments.

PR #674, opened 2015-10-28
Fix gevent._ssl3.SSLContext.read to return empty bytes if SSLError occurs.

salqueng:
In Python3, SSLSocket.read should return b'' if SSLError occurs.
(https://github.com/python/cpython/blob/master/Lib/ssl.py#L792)
However, it returns empty str object('') currently. This fix just change the return value to b''.


thedrow:
Can you check why the tests are failing?

jamadden:
LGTM, thanks! 

One test failure is a test that's flaky under load, the other test is one that's flaky when we're doing leak checks. Neither is related to this.

Issue #702, opened 2015-12-14
sslwrap issues have made it back to python 2.7.5 (Red Hat)


GarrettHeel:
After upgrading to the Centos 7.2 release today, I noticed that the following patch has been backported to python 2.7.5:

https://bugzilla.redhat.com/show_bug.cgi?id=1259421
https://rhn.redhat.com/errata/RHSA-2015-2101.html

This was previously causing the `sslwrap` issues in python >= 2.7.9 and is now also affecting 2.7.5.

Relevant links:

Commit that fixed 2.7.9: https://github.com/gevent/gevent/commit/b6fcd5e468a82cf1926760efee3a531e45948d18#diff-117c9127922ead54983bc5b173701483

2.7.9 discussion: https://github.com/gevent/gevent/issues/477

Ivoz:
Can you paste a traceback of what exactly is breaking for you?

This seems mostly because Redhad is creating a Frankenstein 2.7.5 that's not conformant to what either 2.7.5 or 2.7.9 should be. :/

Redhat's EPEL may give you an actual CPython 2.7.9 that works, if I understand what EPEL is for correctly.

jamadden:
Yeah, there was discussion on python-dev. I think RedHat is backporting PEP 466 to their LTS versions. They know that this is going to have compatibility issues and some of those they've planned for (e.g., a config file to switch HTTPS cert verification *back off* by default), but some they can't.

So if that's what's going on here, I'm not sure what to do about. In general, given how closely tied gevent is to the standard library, it doesn't seem reasonable to support frakenversions distributed by particular distros. (For one thing, it could be extremely hard to test; travis doesn't have access to these hybrids.) On the other hand, in this particular case, it is probably possible to do an attr check to decide which version of the ssl module to use, rather than switching explicitly off version.

GarrettHeel:
Thanks for the answers @ivoz and @jamadden.

I don't have the exact traceback since boto is obscuring it but I do see:
```TypeError: __init__() got an unexpected keyword argument 'server_hostname'```

...which appears to be the same as the traceback from https://github.com/gevent/gevent/issues/477

I completely understand that it's difficult for gevent to support these releases and point taken about looking at the EPEL packages instead.

I'm happy to leave it up to you guys to decide whether an attr check is better or to just call it unsupported and leave it at that.

jamadden:
[...]

PR #705, opened 2015-12-16
Use the new ssl module on any python version that natively supports it

jamadden:
Support the new PEP 466 `ssl` interfaces on any Python 2 version that supplies them, not just on the versions it officially shipped with. Some Linux distributions, including RedHat/CentOS and Amazon have backported the changes to older versions. Reported in #702.

Tests pass locally for me on Python 2.6, 2.7.8 and 2.7.11. I don't have access to one of the backported versions for testing.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -7,7 +7,11 @@
 1.1rc3 (Unreleased)
 ===================
 
-- TBD
+- Support the new PEP 466 :mod:`ssl` interfaces on any Python 2
+  version that supplies them, not just on the versions it officially
+  shipped with. Some Linux distributions, including RedHat/CentOS and
+  Amazon have backported the changes to older versions. Reported in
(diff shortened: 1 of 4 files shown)


blakeblackshear:
Beautiful! This officially fixes the downstream issue with wal-e: https://github.com/wal-e/wal-e/issues/214

This is on a fully upgraded CentOS 7.2 install with the Python 2.7.5 frankenstein.

jamadden:
Thanks very much @blakeblackshear! Very pleased it fixes your issue. I'm hoping to get confirmation from one of the other contributors to the original report in #702 (those not experiencing a segfault, just a python-level error) before attempting a merge. As far as I can tell, this isn't especially risk so I probably won't wait *that* long for feedback, but it still may be a little while before there's a new official RC.

Issue #704, opened 2015-12-16
Segfault on PyPy 4.0.1 with gevent 1.1b6 thru 1.1rc2


fasaxc:
I work on Project Calico, an open source networking project that uses gevent for concurrency. Testing on Ubuntu 14.04, Our UT suite worked fine on PyPy 2.6.1+dfsg-2 but after upgrade to PyPy 4.0.1+dfsg-1~ppa1~ubuntu14.04, it segfaults reliably.

Commenting out the test that happens to be running while it segfaults doesn't help, it just segfaults somewhere else instead.

I uploaded a core file here: https://transfer.sh/nFUKg/core.gz

The kernel log points the finger at gevent:
```
[ 7819.365318] nosetests[71653]: segfault at 30 ip 00007f6f6096e7f2 sp 00007fffea134050 error 4 in _semaphore.pypy-26-x86_64-linux-gnu.so[7f6f6096a000+d000]
```

I was testing https://github.com/projectcalico/calico/commit/19dc016686166562fd9a506ec019b41578688e48

If you'd like to re-run the test, you could check out that commit and run 
```./run-unit-tests.sh -e pypy --develop``` 
then, to run without tox in the way: 
```
source .tox/pypy/bin/activate
nosetests -vv calico.felix
```

fasaxc:
Also reported here against PyPy, where they've done some work to crack the coredump: https://bitbucket.org/pypy/pypy/issues/2209/segfault-while-running-our-uts-after

fasaxc:
From the thread there:

> Note that the current status of gevent on PyPy is unclear at this point. There was a cffi version, which has apparently been merged back and is now in gevent's trunk; I think that's what we are using here. However, clearly not everything is in cffi, because I still see this Cython code. It seems that nobody on #pypy is able to comment more precisely about that. Maybe you should report this issue to gevent developers. They might rightly point back to PyPy's CPython compatibility layer (cpyext), but at least we'd know what to look for.

Maybe you could chime in on that thread?  The PyPy devs seem very open to helping with the investigation.

Issue #580, opened 2015-06-18
CFFI bindings for c-ares


thedrow:
It might be nice to provide better DNS resolving engines for PyPy as well.

jamadden:
The existing Cython-based c-ares bindings seem to work fine, as far as the test-suite is concerned. I've enabled them for PyPy (as an option, not as a default, the same for cpython). Note that I haven't attempted any performance comparisons because That Would Be Hard.

thedrow:
CPython extensions are generally slower on PyPy so I think that there is value in creating a CFFI extension for it.
I also don't know how to measure performance of c-ares in PyPy with a C extension and a CFFI extension.
Maybe we could use the vmprof for that?

jamadden:
> CPython extensions are generally slower on PyPy so I think that there is value in creating a CFFI extension for it.

That can certainly be true, depending on how much time is spent interfacing between C and Python. Given that's it's cython-ized code, my guess is "quite a lot". But then again, it's DNS lookups. How often is that actually a bottleneck in a Python program? More to the point, how often is the local interpreter overhead a bottleneck compared to network overhead in DNS resolution?

That said, I wasn't trying to say that a CFFI binding would never be useful (although it does double the maintenance requirements). I was just happy to be able to say that the functionality does exist and can be used, even if it doesn't perform optimally.

jamadden:
> I wasn't trying to say that a CFFI binding would never be useful (although it does double the maintenance requirements)

Hmm. After having gotten the corecffi loop running on CPython, it appears there's very little performance penalty relative to the Cython loop (~5-10%, depending on platform). If that turns out to hold true (and it looks like it should) then I would be in favor of dropping the Cython c-ares bindings in favor of CFFI bindings that are usable and stable on both platforms (assuming we don't run into any windows-related issues, as I did with the CFFI loop). (That's certainly where I intend to start investigating a libuv loop :) )

thedrow:
Sounds like a plan.

jamadden:
#774 is probably a bit of a blocker for this. And as a side note, a libuv loop could dispense with c-ares altogether in favor of the async DNS builtin to libuv.

thedrow:
Are there plans to move from libev to libuv?

jamadden:
I wouldn't call them exactly plans at this point, but it's something I want to take a serious look at.

jgehrcke:
IMO one of the nicest things libuv would bring along is IOCP support.
Also see
[...]

PR #708, opened 2015-12-18
Use a manually locked pure-Python Semaphore implementation on PyPy

jamadden:
Issue #704 and the ensuing [discussion on the PyPy bugtracker](https://bitbucket.org/pypy/pypy/issues/2209/segfault-while-running-our-uts-after) detail an *extremely* easy-to-reproduce crash involving the interaction between Cython compiled objects (in general, any object using the cpyext api) and the PyPy garbage collector. Because of that and other issues we've encountered with PyPy and the cpyext API, such as memory leaks, this PR proposes that we stop compiling the Semaphore with cython and go back to a pure-python implementation.

What are the implications?

First, raw speed. Under CPython presumably Cython is an improvement, but under PyPy it is not. In testing `sem.acquire();sem.release()` with `timeit`, the Cython version clocks in at 5usec per iteration, while the version we end up with here comes in at 0.262usec, a substantial improvement (and of the same order of magnitude as the CPython+Cython time of 0.141usec). So we're good there.

Second, thread safety. Because the Cython compiled methods are called with the GIL held, and never release them, each method was implicitly threadsafe and automatically non-reentrant. There's never an opportunity for a thread switch to occur while a Cython method is running until it calls into a pure-python method or returns. (The semaphore does make some calls to `get_hub` which would allow a thread switch but only when we're in a stable state.) As for greenlets on the same thread, the semaphore carefully manages its switches.

If we don't compile with Cython, we need to do something else to be thread safe. This implementation wraps a native thread lock around the critical methods that manipulate the counter: `wait`, `acquire` and `release`, being careful to drop the lock when we go to sleep in the hub. Although it would be nice use a single lock instead of one per instance to both match the GIL semantics and keep overhead as low as possible, we can't do that (see below) (note that in the timeit tests, the per-instance lock and context manager system in use now performs about well as a simpler global strategy).

I *suspect*, but haven't confirmed, that there may be a race condition with signal handlers trying to acquire a Semaphore at the same time that normal code tries to acquire it, leading to a deadlock. That seems unlikely to me, though.

Finally, `test__threading_vs_settrace` and its original version in the python test suite, `test_threading:test_finalize_with_trace`. Both of them ensure that a tracing function that accesses `threading.current_thread()` during interpreter shutdown do not hang. This refers to a bug in the interpreter's threading implementation that was fixed for 2.6 (and so doesn't really concern us). The gevent test case goes further to say that no tracing function should ever see the workings of Semaphore, which is easy to ensure with Cython, and pretty much impossible otherwise.

I've scratched my head over that quite a bit and tried various recursive-lock-like strategies to minimize any interference that a tracing function might have, but I don't think it's possible to be perfect. The current solution works as expected on tracing a lock function when the trace function uses no locks or when the trace function uses a different lock. When the trace function tries to acquire or release the same lock that is being traced, a LoopExit is raised; that seemed better than the alternative which was to deadlock. (Ack, just realized that the last change to make that work probably broke the multi-threaded semantics so I need to add a new commit, but I'm opening this PR now for review.)

Comments, feedback, alternate ideas very welcome!


jamadden:
I think I fixed the threading reentrancy problem. It slows down the semaphore to 0.5usec, but otherwise all the tests pass.

thedrow:
Can't we implement the lock in C using CFFI?

thedrow:
Also, I think we're discovering more and more that we need to implement #580 or disable c-ares on PyPy just to avoid crashes.

jamadden:
> Can't we implement the lock in C using CFFI?

Indeed, that was my original intent (thus the name of the branch). That turns out to be rather ugly and more involved than this approach though, I think. The entirety of `acquire`, `wait` and `release` have to be implemented in C, making callbacks to Python (well, I say "entirety", but there still has to be a thin Python wrapper that sets up the call into C). There has to be an additional C struct containing an integer to hold the value that's manipulated in C, and the value synced with what's visible from the Python side.

The Semaphore class is relatively complex and widely used, and I don't want a completely separate implementation of it, at least not this late in the game, so that all has to be accomplished by monkey-patches, similar to what's being done in this PR.

Finally, while the C functions would be suitably atomic, it's not clear if they would hold the GIL or not, meaning we may still need a lock. C function calls are [documented as *dropping* the GIL](http://cffi.readthedocs.org/en/latest/using.html) but that may only apply to function pointers; it's not totally clear to me.

jamadden:
>  disable c-ares on PyPy just to avoid crashes.

I think the documentation is pretty clear about the possible risks of using the ares resolver on PyPy. Because it's optional and not on by default, I don't see a lot of benefit in removing the capability entirely; indeed, using Semaphore through Cython has led to the detection of a number of PyPy bugs---they can't know about bugs in code that's not being used.

I also think that the ares resolver has a much different usage pattern than Semaphores do. A typical program will create exactly *one* instance of a Cython-compiled class that lives for the duration of the program so the GC interactions should be eliminated. Clients should never actually be dealing with any of the Cython-compiled code, only our wrapper in resolver_ares.py so we're free to make adjustments/workarounds there as necessary.

PR #711, opened 2015-12-24
compatible compile on smartos/solaris2.11

cc272309126:
due to the statfs is different on smartos or solaris, so the gevent can'be compile. I check the configure to fix this question.

### DIFF
--- a/libev/configure
+++ b/libev/configure
@@ -11481,8 +11481,11 @@ CC="$lt_save_CC"
 
 
 
-
-for ac_header in sys/inotify.h sys/epoll.h sys/event.h port.h poll.h sys/select.h sys/eventfd.h sys/signalfd.h
+ck_headers=('sys/inotify.h' 'sys/epoll.h sys/event.h' 'port.h poll.h sys/select.h' 'sys/eventfd.h' 'sys/signalfd.h')
+if [[ "$host_os" =~ "^solaris" ]]; then
+    ck_headers=('sys/epoll.h sys/event.h' 'port.h poll.h sys/select.h' 'sys/eventfd.h' 'sys/signalfd.h')
+fi
+for ac_header in ${ch_headers[@]}
 do :
   as_ac_Header=`$as_echo "ac_cv_header_$ac_header" | $as_tr_sh`
 ac_fn_c_check_header_mongrel "$LINENO" "$ac_header" "$as_ac_Header" "$ac_includes_default"
@@ -11492,11 +11495,14 @@ if eval test \"x\$"$as_ac_Header"\" = x"yes"; then :
 _ACEOF
 
 fi
-
 done
 
 
-for ac_func in inotify_init epoll_ctl kqueue port_create poll select eventfd signalfd
+_func=('inotify_init' 'epoll_ctl' 'kqueue' 'port_create' 'poll' 'select' 'eventfd' 'signalfd')
+if [[ "$host_os" =~ "^solaris" ]]; then
+    ck_func=('epoll_ctl' 'kqueue' 'port_create' 'poll' 'select' 'eventfd' 'signalfd')
+fi
+for ac_func in ${ck_func[@]}
 do :
   as_ac_var=`$as_echo "ac_cv_func_$ac_func" | $as_tr_sh`
 ac_fn_c_check_func "$LINENO" "$ac_func" "$as_ac_var"


jamadden:
This is something that needs to be addressed upstream with the libev maintainers. `configure` is not maintained by hand, it is automatically built by the autotools package, and without much customization. Is it possible that autotools (autoconf) doesn't support your operating system yet? What's the full version of the operating system you're using and the compiler tool chain? Can you provide the failed `configure` output on your system?

cc272309126:
Hi, I use smartos to compile the gevent, the version is SunOS joyent_20151210T194528Z i86pc i386 i86pc Solaris. it is a branch of solaris.
the configure is run correctly on the my os.  when configure run, it will produce makefile. and produce HAVE_INOTIFY_INIT. this option will lead the libev compile <sys/stat.h>, in smartos this header file has different statfs declaration。so this lead gevent can‘t be compile。

cc272309126:
on solaris statfs is like this  int statfs(const char *, struct statfs *, int, int); and struct statfs like "
struct  statfs {
        short   f_fstyp;        /* File system type */
        long    f_bsize;        /* Block size */
        long    f_frsize;       /* Fragment size (if supported) */
        long    f_blocks;       /* Total number of blocks on file system */
        long    f_bfree;        /* Total number of free blocks */
        ino_t   f_files;        /* Total number of file nodes (inodes) */
        ino_t   f_ffree;        /* Total number of free file nodes */
        char    f_fname[6];     /* Volume name */
        char    f_fpack[6];     /* Pack name */
};
"

jamadden:
Does it work for you if you define `EV_STANDALONE` first? E.g., `$ CFLAGS="-DEV_STANDALONE" pip install --pre gevent`?

jamadden:
Another option would be to try this patch to `ev.c`. If it works correctly, it should disable inotify on Solaris:

```
diff --git a/libev/ev.c b/libev/ev.c
index 364262e..43488f3 100644
--- a/libev/ev.c
+++ b/libev/ev.c
@@ -127,6 +127,7 @@
 # endif

 # if HAVE_PORT_H && HAVE_PORT_CREATE
+#  define EV_USE_INOTIFY 0
 #  ifndef EV_USE_PORT
 #   define EV_USE_PORT EV_FEATURE_BACKENDS
 #  endif
```

doublerebel:
Thanks for these clues, this PR helped me compile libev for ArangoDB on SmartOS.

However, instead of making this change in configure, it's possible to use these defines as compiler flags:
```sh
-D HAVE_PORT_H=1 -D EV_USE_PORT=1 -D EV_USE_INOTIFY=0 -D EV_USE_EVENTFD=0
```
Also if libev is used in a threaded environment, `-D _REENTRANT=1` is also necessary.  I'm using all these flags to successfully compile, without changing the `configure`.

There is a relevant open bug for inotify and eventfd on SmartOS, **[OS-4419: inotify and eventfd attach routines don't properly handle several edge cases](https://smartos.org/bugview/OS-4419)**.

I don't blame @cc272309126 for going after the configure as the first fix, it took me a while with gdb to figure out what defines were needed in the chain of libev macros.  The libev docs are a bit lacking for Solaris and don't mention all the flags.

Ivoz:
What's a nice reliable way to tell you're on SmartOS in shell script? Maybe something could be put in `Makefile` or similar to set those.

jamadden:
I posted about this to the libev mailing list: this is really a libev problem and ideally it would be fixed there; it's hard supporting extra platforms that upstream doesn't support because this will just keep happening (or worse, misbehave more subtly than a compile error). But I'm not a Solaris user, so my post was speculative and missed some of the necessary points. Maybe one of the solaris users could try posting to libev again with a working solution? That might carry more weight.

jamadden:
Although there's been no motion on this from the libev devs, I'm unwilling to accept a change to the upstream configure script (for example, it changed already when updating our bundled version of libev). I would review a different PR that detects SmartOS, and [as mentioned](https://github.com/gevent/gevent/pull/711#issuecomment-188647053) adds [the appropriate defines](https://github.com/gevent/gevent/pull/711#issuecomment-174749902) to `Makefile` (or, more likely, to `setup.py`).

In the meantime, this issue is [linked to] from the gevent docs so the defines won't get lost.

Thanks everyone for looking into this issue!

PR #712, opened 2015-12-24
Use latin-1 as decoding in unquote for conformance with PEP 3333

rubendv:
In Python 3, unquote will decode the resulting string according to the encoding keyword, which is utf-8 by default. However, according to PEP 3333 the native string representing PATH_INFO (among others) can only contain unicode codepoints from 0 to 255, which is why we need to decode to latin-1 instead of utf-8 here. Frameworks like Django turn this back into bytes for proper UTF-8 decoding by first encoding using latin-1 encoding, which can raise a UnicodeEncodeError if the string was decoded with UTF-8.

After applying this fix, gevent behaves the same as wsgiref for our application.

In Python 2 there is no such keyword so there will need to be a check for Python 3 here as well.

### DIFF
--- a/gevent/pywsgi.py
+++ b/gevent/pywsgi.py
@@ -949,7 +949,7 @@ class WSGIHandler(object):
             path, query = self.path.split('?', 1)
         else:
             path, query = self.path, ''
-        env['PATH_INFO'] = unquote(path)
+        env['PATH_INFO'] = unquote(path, encoding='latin-1')
         env['QUERY_STRING'] = query
 
         if self.headers.typeheader is not None:
--- a/gevent/pywsgi.py
+++ b/gevent/pywsgi.py
@@ -16,6 +16,7 @@ import sys
 import time
 import traceback
 from datetime import datetime
+
 try:
     from urllib import unquote
 except ImportError:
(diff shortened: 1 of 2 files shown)


jamadden:
So that broke all the builds. On Python 2, the initial problem is the lack of an 'encoding' keyword, so that's the first thing that needs to be fixed; I would probably consider using a `functools.partial` instance at module scope under Python 3 to avoid any runtime overhead of an if check on Python 2.

Then the international encoding tests are broken under Python 3. Those need to be looked into closely; I'm afraid we may be looking at an effective functionality regression if we change the encoding (even if it does adhere more closely to the spec). That would tend to put this in the category of "post 1.1 release" changes at this point.

rubendv:
Do you have a preferred way of checking for python 2 or 3? I see you are not using six.

jamadden:
`gevent.hub.PY3`, etc. This is already imported in `pywsgi`.

rubendv:
Those ssl test failures look unrelated to me?

jamadden:
Yeah, it looks like `svn.python.org` has a new certificate.

jamadden:
The test_ssl issues are fixed in master if you want to rebase.

rubendv:
I was able to fix the failing international encoding test.

rubendv:
Since 1.1 is the first version of gevent to support python 3 (except for the RCs of course), wouldn't it be better to include this fix? This issue does not exist under python 2, so there would be no regressions for people who don't use release candidates.

jamadden:
Having looked into it more closely, I do agree that this is an appropriate change for 1.1rc3. I've also verified that this produces the same `environ` dict that we'd get under `waitress`. Thanks!

PR #724, opened 2016-01-24
Demonstrate how to install gevent 1.1rc* with pip

Ivoz:
Putting this front and center in the readme might help OS X users, or even others to easily install gevent 1.1 rc's.


jamadden:
LGTM.

PR #725, opened 2016-01-24
Add .tox to gitignore

Ivoz:


jamadden:
Ahh, I had `.tox` in my global configuration from way back. LGTM.

jamadden:
That travis failure was one of the ones that can very occasionally be flaky under load, and obviously could not have been caused by this.

PR #729, opened 2016-01-28
TODO is no longer being currently used or needed

Ivoz:


jamadden:
Maybe I'm old-school, but I like for projects to have an on-disk/in-sdist document of their issues, even if it's a pointer to somewhere else (or, especially if if's a pointer). To me, that goes along with having a README and a NEWS (or ChangeLog).

PR #733, opened 2016-02-07
Note Jason Madden as current maintainer

Ivoz:
@jamadden I think this is pretty well-deserved by now, but I'll leave to you whether you agree or wish not to have your name up for any reason.

:+1:


jamadden:
I'd be honored, but maybe @denik has some input.

denik:
+1 Absolutely, long overdue.

Issue #685, opened 2015-11-06
syntax error in gevent/_socket3.py  with python2


jlec:
```python
writing byte-compilation script '/var/tmp/portage/dev-python/gevent-1.1_beta6/temp/tmpOh53QF.py'
/usr/bin/python2.7 -OO /var/tmp/portage/dev-python/gevent-1.1_beta6/temp/tmpOh53QF.py
  File "/usr/lib64/python2.7/site-packages/gevent/_socket3.py", line 165
    def makefile(self, mode="r", buffering=None, *,
                                                  ^
SyntaxError: invalid syntax

removing /var/tmp/portage/dev-python/gevent-1.1_beta6/temp/tmpOh53QF.py
```

jamadden:
This is known and expected and generally not a problem. See e.g., #601 among others.

thedolphin:
Insert
`%define _python_bytecompile_errors_terminate_build 0`
in your rpm spec

thedolphin:
insert
`%define _python_bytecompile_errors_terminate_build 0`
in your spec file

Issue #601, opened 2015-07-02
gevent 1.1a1 pip package contains a syntax error


the80srobot:
On line 157 of _socket3.py:
    def makefile(self, mode="r", buffering=None, *, encoding=None, errors=None, newline=None):

Notice the asterisk.

jamadden:
That is [correct syntax under Python 3](https://docs.python.org/3/library/socket.html#socket.socket.makefile), and that file is only used under python 3. You may see warnings from setuptools under Python 2 when installing the package because setuptools tries to byte-compile *all* .py files, regardless of whether or not they are used.

jgehrcke:
I understand that this is no issue technically, but we should anticipate this to be reported often in the future if it stays as it currently is. The `SyntaxError: invalid syntax` message appears rather prominently by the end of the installation log and will probably irritate a number of people.

Which options do we have to prevent this from happening?

jamadden:
No options I'm aware of. Note that a large number of prominent projects that straddle 2/3 (e.g., gunicorn) also produce SyntaxErrors when installed on one platform or another; a google search turns up lots of hits. Supposedly this is fixed in pip (pypa/pip#1954)...

JokerQyou:
Actually this is preventing us from packaging any scripts that use gevent 1.1 branch modules, because pyinstaller will break on this syntax issue.

jgehrcke:
Also removed:

I was supported with a PhD fellowship by the [Studienstiftung des 
deutschen Volkes](http://en.wikipedia.org/wiki/Studienstiftung) and 
conducted my PhD in the framework of the [Dresden International PhD 
Program](http://www.dresden-ipp.de/). I regularly spoke on conferences 
and published in peer-reviewed journals.

pigmej:
For any packagers: just remove conflicting files (like _socket3.py) on python2 environments in some preprocess script. Or use pip >= 7.1.2

JokerQyou:
FYI, update PyInstaller will fix packaging with gevent 1.1. :)

Issue #478, opened 2014-09-12
Setup script from Git version breaks during "install"


slaff:
When I try to install gevent from the latest source from the git  repository using the command below:

```sudo python setup.py install```

I get the following error message:

```
Extracting gevent-1.1-py2.7-linux-i686.egg to /usr/local/lib/python2.7/dist-packages
  File "/usr/local/lib/python2.7/dist-packages/gevent-1.1-py2.7-linux-i686.egg/gevent/_socket3.py", line 133
    def makefile(self, mode="r", buffering=None, *,
                                                  ^
SyntaxError: invalid syntax
```

My Python version is:

```
Python 2.7.6 (default, Mar 22 2014, 22:59:38) 
[GCC 4.8.2] on linux2
```

**Comment by thedrow** [comment]
This issue is not relevant any more since Gevent master can be installed on Python 2 correctly now.

**Comment by oblalex** [comment]
Hi!
I have no idea why, but this issue is still relevant with 1.1b6 @ Python 2.7.
Any thoughts?

**Comment by jamadden** [comment]
That SyntaxError is expected and *not* a problem. You don't see it using 'pip' to install the distributions from PyPI, but there's nothing to do about it when running setup.py manually. See also #601.

**Comment by oblalex** [comment]
Usage of `pip` works like a charm out-of-the-box. But for a one insane reason I need to get a `RPM` package from `gevent` via `rpmbuild` without usage of `pip`. I cannot change this requirement and I would like to know whether there is some workaround?

Here is a `spec` for the package:

```
Name:           python-gevent
Version:        1.1b6
Release:        1%{?dist}
Summary:        Coroutine-based network library

Group:          Development/Libraries
License:        BSD
URL:            https://pypi.python.org/pypi/gevent
Source0:        https://pypi.python.org/packages/source/g/gevent/gevent-1.1b6.tar.gz
BuildArch:      x86_64

BuildRequires:  python2-devel
BuildRequires:  python-setuptools
BuildRequires:  python-nose
BuildRequires:  pytest
Requires:       python2

%description
Coroutine-based network library

%prep
%setup -q -n gevent-1.1b6

%build
%{__python2} setup.py build

%install
%{__python2} setup.py install -O1 --root %{buildroot}

%files
%doc LICENSE README.rst
/usr/lib/debug/*
/usr/lib64/python2.7/*
/usr/src/*
```

Please, could you share any thoughts on that? I appreciate your help anyway.

jgehrcke:
[...]

PR #740, opened 2016-02-17
Fix a possible typo

cokelaer:
The "def makefile" prototype in gevent/_socket3.py seems wrong with an extra argument called '*'.

This syntax prevents a compilation of the library as a conda recipe.

All tests passed succesfully locally under Python2.7 (running the tests in ./greentest directory).

### DIFF
--- a/gevent/_socket3.py
+++ b/gevent/_socket3.py
@@ -173,7 +173,7 @@ class socket(object):
             sock.setblocking(True)
         return sock, addr
 
-    def makefile(self, mode="r", buffering=None, *,
+    def makefile(self, mode="r", buffering=None, 
                  encoding=None, errors=None, newline=None):
         """makefile(...) -> an I/O stream connected to the socket
 


jamadden:
Thank you, but that's the correct syntax [for Python 3](https://docs.python.org/3/library/socket.html#socket.socket.makefile). See #685, #601, etc.

cokelaer:
I was trying to compile gevent  for bioconda (a set of conda recipes for biology) and since conda provide gevent with a version 1.0.2 (python2.7) and thought that i would create my own conda recipe for python3.5 and latest version of gevent. Yet, I ran into an syntax error:

    File "/anaconda/envs/_build/lib/python2.7/site-packages/gevent-1.1rc4-py2.7-linux-x86_64.egg/gevent/_socket3.py", line 176
    def makefile(self, mode="r", buffering=None, *,

I'm realising that the underlying python version used is actually python 2.7 , which may explain the syntax error.

thanks for the quick reply.

jamadden:
Ok, I'm going to go ahead and close this then. (You may find the comments in #478 to be of further help as it's a similar packaging issue.) Please feel free to open other issues if necessary.

cokelaer:
I've read the different comments from the issues you provided. 
I understand that this is correct syntax under Python3 and indeed it does work under PYthon2.x as well.
However, FYI, it may be an issue with conda. After creating a conda recipe for gevent and running this command:

conda build recipes/gevent/ --python=2.7

The package is compiled but because of the error, a status different from 0 is returned and may have side effects (using bioconda, they use a docker to check that builts are passed correctly and in that case gevent does not).

jamadden:
For downstream distributors with tooling that can't cope with projects that span Python 2 and 3, it's not uncommon for them to modify the source tarball and produce completely separate builds for 2 and 3. That was suggested for the RPM packager; you might consider that. Then, the RPM packager was able to make the tooling work by not precompiling all source files, which would be ideal. (Of course RPM is still Python version specific so it doesn't make much difference.)

Heck, some downstream distributors go so far as to remove libev completely and link to the system-provided libev dynamically. That makes it much harder for upstream to support but they deem it necessary on their platform.

cokelaer:
thanks.

PR #758, opened 2016-03-04
Issue some warnings for patch_all with invalid parameters

jamadden:
The main purpose of this patch is to issue a warning for some errors seen in production cases (running gunicorn apps). I consider them in scope for 1.1 because they're mostly about refining features that were added in 1.1 and which could trip up some unsuspecting users coming from 1.0.

- Patching signal but not os results in SIGCHLD handlers not getting called. Under gunicorn, this leads to the arbiter process not being aware of dead children, causing extended shutdown times and/or out-of-service scenarios. This may be an error in the future (because it's effectively a broken case), but right now it's just a warning because it's possible to not use the broken features.
- Patching multiple times with different arguments has an unintuitive effect. It's not always clear where this is being done either, so issue a warning about it.

Coming from a 1.0 code-base that had `os=False` in one monkey-patch call but not another (two semi-independent modules glued together in the app), the extended shutdowns were annoying to debug.

I also fixed some threading attributes being reported incorrectly from get_original that I happened to notice.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -30,6 +30,16 @@
   requires having Cython installed first. (Note that the binary installation
   formats (wheels, exes, msis) are preferred on Windows.) Reported in
   :issue:`757` by Ned Batchelder.
+- Issue a warning when :func:`~gevent.monkey.patch_all` is called with
+  ``os`` set to False (*not* the default) but ``signal`` is still True
+  (the default). This combination of parameters will cause signal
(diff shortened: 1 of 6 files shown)


jamadden:
The last AppVeyor build had at least all the 64-bit builds green, but we're into a busy time now so parts have started to run into the timing flakiness.

jamadden:
Nope, looks like just the first one hit that timing snag. Merging.

PR #762, opened 2016-03-07
Fix build with Python 3

sunpoet:


jamadden:
That file is not used on Python 3. Older versions of setuptools/pip may emit a *warning* that it can't be byte-compiled during installation, but it is not a runtime problem.

PR #766, opened 2016-03-09
Drop support for python 2.6.

jamadden:
On travis, start a build with 2.7.8 so that _ssl2.py gets
tested (previously tested by Python 2.6). Do so with the test_ssl.py
from that version (because the changes are large), but keep using the
2.7.11 tests with just a few removals otherwise.


jamadden:
As Brett Cannon [eloquently puts it](http://www.snarky.ca/stop-using-python-2-6), it's time to move on. The Zope project is another project I know of that's dropping support for 2.6

Does anyone have any issues with this, or reasons to keep 2.6 support around? @ivoz @denik

Ivoz:
I would suggest announcing coming drop of support in 1.1.1 release notes (for instance, you could release 1.1.1 as simply update of bundled libraries if there is nothing else much to fix) and then actually dropping support in a 1.2.

Otherwise folks can tend to get irrationally angry. Gevent being a 2-only library for a long time, I wouldn't be surprised if there were still a lot of 2.6 users.

jamadden:
That's a good point. I had meant to mention that in the original 1.1.0 announcement, but I forgot. It's easy enough to add to PyPI and whatsnew_1_1.html, which I'll do, but that's not enough.

I wasn't planning to update the bundled libraries in the 1.1.x branch: there are some changes in c-ares that we had to adapt to that may make it impossible to use non-embedded c-ares versions that are too old (maybe). System packagers tend to disable all our embedding, sadly, so that could be too big a change for a 1.1.x release.

But surely someone will find a bug that will motivate a 1.1.1 :)

renstrom:
Might be a good idea to add a `DeprecationWarning` to `setup.py` for people installing on Python 2.6?

jamadden:
I'm not sure that trying to print a warning from the 1.1.x `setup.py` on 2.6 would be that helpful, simply because there's no way to respond to it (other than upgrading the entire Python installation). It's not like the 1.1.x branch is going to suddenly stop working on Python 2.6, so everytime a 2.6 user installs 1.1.1, that warning doesn't really tell them anything...it just gets old and annoying after awhile. (I think this is the same reason the DeprecationWarnings for Python 3 stuff aren't shown in Python 2.7 by default.)

Or were you suggesting that `setup.py` in 1.2/master should throw an exception if it gets run under 2.6, like we do for other unsupported configurations (E.g., PyPy on Windows)? That does seem potentially useful to me.

PR #768, opened 2016-03-10
Update c-ares to 1.11.0

jamadden:
- Replaced the whole directory with current source. This adds a few more
  scripts that we could remove again if desired, but it's easier just to
  be able to copy everything.
- However, the new test directory requires c++11, which travis doesn't
  have, so it's ditabled.
- The invocation of configure had to lose the CONFIG_FILES= param,
  because with that set, configure would fail to create the Makefile
  which in turn caused the whole thing to fail.


jamadden:
The travis build is green, as was an earlier appveyor build. Note that appveyor doesn't actually run the full suite of resolver tests with ares, just the minimum that are specific to the ares resolver. Merging.

PR #769, opened 2016-03-11
Switch to using prospector for lint checks

jamadden:
And landscape.io for PR reviews.

I can squash this into fewer commits, just opening now for comparisons and discussion on the ruleset. (I'm all for pretty strict rules, BTW, why have them otherwise?)

This gets us to 100% passing on Python 2 and Python 3. There are a few of the warnings that should be addressed in individual commits going forward, but at least now the problem areas are easily marked and grepable.

There were a few minor issues that this revealed and a few important ones; thanks pylint! The important ones are in the changelog. But mostly this was a case of reviewing each error one-by-one and deciding if it needed to be changed now (or could be changed without breaking BWC), and if not, disabling that particular case. The goal would be to keep from adding now errors/disable-comments in the future, and to gradually whittle away at the ones that exist.


jamadden:
I squashed the commits into a smaller set.

jamadden:
I think I need to move the pep8 configuration from setup.cfg over to .landscape.yml; this assists in running prospector from, e.g., emacs flymake/flycheck mode.

While doing so I'll comment and review each of the pep8 rules.

jamadden:
FWIW, here's a barebones Emacs flycheck configuration that works with current prospector version (0.11.7) and emacs 24.5:

```lisp
(setq jm-flycheck-prospector-pylint ".pylintrc")
(setq jm-flycheck-prospector-profile ".landscape.yml")

(flycheck-define-checker python-prospector
  "A Python syntax and style checker using Prospector.
See URL `http://prospector.readthedocs.org/en/latest/index.html'."
  :command ("prospector"
            "-M"
            "-o" "emacs"
            (config-file "-P" jm-flycheck-prospector-profile)
            (config-file "--pylint-config-file" jm-flycheck-prospector-pylint)
            source-inplace
            )
  :error-patterns
  ((error line-start
          (file-name) ":" (one-or-more not-newline) (optional "\r") "\n"
          (one-or-more " ") "L" line ":" column
          (message (minimal-match (one-or-more not-newline))  (optional "\r") "\n"
                   (one-or-more not-newline)) (optional "\r") "\n" line-end)
   )
  :modes python-mode)

(add-to-list 'flycheck-checkers 'python-prospector)
```

jamadden:
Builds are green. I'm happy enough with the initial ruleset; it can be narrowed and cleanup up in subsequent PRs. I'm going to go ahead and merge.

PR #770, opened 2016-03-12
Fixed typo error

mayflaver:


jamadden:
Thanks. We'll ignore the landscape check until #769 lands.

PR #773, opened 2016-03-16
Make socket.recvmsg() and socket.sendmsg() cooperative.

jceel:


jamadden:
Thank you for the contribution! A few comments/requests:

- Can you rebase on latest master? That should clear up any spurious landscape.io failures.
- I think you need to add these methods conditionally, not all platform support them, and the test suite relies on detecting the attribute on the class to know whether or not to run certain tests. If gevent unconditionally provides them, we'll get test failures on certain platforms. Something like:
```python
if hasattr(_socket.socket, 'sendmsg'):
   def sendmsg(self, ...):
       ...
   def recvmsg(self, ...):
       ...
```
- Can you add an implementation for `recvmsg_into`?
- And last, we'll need to figure out why the stdlib tests of these functions are failing now that these methods are present and how to fix them.

renstrom:
Just a side-note, I think `hasattr()` is considered bad or unpredictable on Python 2.

https://hynek.me/articles/hasattr/

jamadden:
> Just a side-note, I think hasattr() is considered bad or unpredictable on Python 2.

Sure, in some cases the semantics may not be exactly what you want and you have to be careful, but (a) this is a *class* (not an *instance* of a class) from the standard library, both of which mean the behaviour is known and predictable, and (b) this is Python 3 :)

renstrom:
>(b) this is Python 3 :)

I need to look closer next time :astonished:

jamadden:
I'm going to go ahead and see if I can diagnose the failures.

jamadden:
Thanks again for your contribution. The remaining errors turned out to be pretty simple to solve, and I've merged your commit.

PR #779, opened 2016-03-22
Extend pywsgi user defined error logging

sean-peters-au:
Extend the user provided error logger to replace all usages of handle_error.

An example of why this might be needed:

handle_error ends up printing to stderr environment details. These can include sensitive data such as in the HTTP_AUTHORIZATION header.

### DIFF
--- a/gevent/pywsgi.py
+++ b/gevent/pywsgi.py
@@ -897,7 +897,6 @@ class WSGIHandler(object):
 
         try:
             try:
-                raise Exception
                 self.run_application()
             finally:
                 close = getattr(self.result, 'close', None)
@@ -942,7 +941,7 @@ class WSGIHandler(object):
 
     def _handle_server_error(self, ex):
         type, value, tb = sys.exc_info()
-        self.log_error('Server response error: %s failed with %s\n%s\n\n' % (self.environ, getattr(type, '__name__', 'exception'), tb))
+        self.log_error('Server response error\nContext %s\n failed with %s' % (self.environ, traceback.format_exc()))
         self._send_error_response_if_possible(500)
 
     def _handle_client_error(self, ex):
--- a/gevent/pywsgi.py
+++ b/gevent/pywsgi.py
@@ -654,10 +654,6 @@ class WSGIHandler(object):
             if not self.read_request(self.requestline):
                 return ('400', _BAD_REQUEST_RESPONSE)
         except Exception as ex: # pylint:disable=broad-except
-            # Notice we don't use self.handle_error because it reports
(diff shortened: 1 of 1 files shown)


jamadden:
Passing errors to the loop's method is indeed quite deliberate. This removes that. I don't think we can accept that change.

sean-peters-au:
Can you suggest a different change to stop the above example from occurring? The user could provide headers they don't want logged when constructing their pywsgi server. I thought the submitted solution was less hacky than this approach though.

jamadden:
Sending context and errors through the loop's `handle_error` function is the way gevent handles these things. Therefore, to change these things, you can either change (set) the loop's handle_error function in your program, or you can subclass the gevent-provided WSGIHandler and override *its* `handle_error` function.

sean-peters-au:
Okay. If this doesn't merge I'll subclass the WSGIHandler and override handle_error. :)

I must say though that as a user I had expected that this was exactly what the error_log parameter in the WSGIServer would do. The description in the docstring seems to indicate so.

Perhaps by default errors should go through handle_error else if error_log is set then through to error_log?

jamadden:
> I must say though that as a user I had expected that this was exactly what the error_log parameter in the WSGIServer would do. The description in the docstring seems to indicate so.

I can understand the confusion. The "you" in the docstring refers to the WSGI application and the `wsgi.errors` parameter of the WSGI spec---expected errors, if you will. On the other hand, If you're getting to the server's `handle_error` method, then something has gone dramatically wrong; an unexpected error that neither the application nor the server was prepared for. In an application out of development those should be vanishingly rare (hopefully); most frameworks, for example, include some sort of error middleware that catches exceptions that bubble up from the lower WSGI application and transforms them into appropriate error responses after having done its own logging.

sean-peters-au:
We noticed these occurring before ECONN_RESET was silenced in gevent 1.1.0. I'm not sure how errors could occur now, but regardless in our example of HTTP_AUTHORIZATION, we don't want to take any risks of logging sensitive data.

jamadden:
There is precedent in the WSGI space for using `dicts` that don't print their details by default. Two examples I can think of: [repoze.who](github.com/repoze/repoze.who) uses an `Identity` in the environ that prints *nothing* by default, and [webob's](github.com/pylons/webob) (and hence Pyramid's) `MultiDict` (which is used extensively) takes a *blacklist* approach and prints everything except selected values.

I'm kinda liking the idea of using a custom `dict` subclass for `environ` for this and possibly other reasons. I may give that a try to see what it looks like...

PR #780, opened 2016-03-22
Improve compliance of select.select and select.poll

jamadden:
- Compliance: If :func:`gevent.select.select` is given a negative *timeout*
  argument, raise an exception like the standard library does.
- Compliance: If :func:`gevent.select.select` is given closed or invalid
  file descriptors in any of its lists, raise the appropriate
  ``EBADF`` exception like the standard library does. Previously,
  libev would tend to return the descriptor as ready. In the worst
  case, this adds an extra system call, but may also reduce latency if
  descriptors are ready at the time of entry.
- Compliance: :meth:`gevent.select.poll.unregister` raises an exception if *fd* is not
  registered, like the standard library.
- Compliance: :meth:`gevent.select.poll.poll` returns an event with
  ``POLLNVAL`` for registered fds that are invalid. Previously it
  would tend to report both read and write events.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -33,6 +33,20 @@
 - The modules :mod:`gevent.os`, :mod:`gevent.signal` and
   :mod:`gevent.select` export all the attributes from their
   corresponding standard library counterpart.
+- Compliance: If :func:`gevent.select.select` is given a negative *timeout*
+  argument, raise an exception like the standard library does.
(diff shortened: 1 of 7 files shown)


jamadden:
The appveyor builds are green. I have no clear idea why prospector is reporting new errors, those files were unchanged. I see that versions of lint tools have been updated but I'm not sure what exactly changed. I'll fix those errors on a different branch.

PR #781, opened 2016-03-24
Security: Use subclass of dict for WSGI environ.

jamadden:
Right now, this is only used for suppressing the printing of potentially sensitive information, but in the future there could be other uses.

This is technically not compliant with PEP3333 which specifies that the type(environ) must be dict, but it's not clear if that practically matters anymore (it looks like it might be a holdover from supporting
Python 1.5.2 before one could subclass the builtin dict; see https://mail.python.org/pipermail/web-sig/2003-December/000394.html).

Discussion on web-SIG ongoing at: https://mail.python.org/pipermail/web-sig/2016-March/005455.html Any comments here? 

Fixes #779.

### DIFF
--- a/changelog.rst
+++ b/changelog.rst
@@ -53,6 +53,11 @@
 - Servers: Default to AF_INET6 when binding to all addresses (e.g.,
   ""). This supports both IPv4 and IPv6 connections (except on
   Windows). Original change in :pr:`495` by Felix Kaiser.
+- Security: The pywsgi ``environ`` dict doesn't print its contents by
+  default anymore, which could have lead to potential secure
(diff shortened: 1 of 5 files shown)


sean-peters-au:
Cool! Thanks!

jamadden:
Unfortunately, it turns out that being secure-by-default [breaks WebOb and Pyramid](https://mail.python.org/pipermail/web-sig/2016-March/005459.html), so we'll have to take a slightly different tactic. gevent can still protect itself, but applications running within it would be subject to their own inadvertent information disclosure vulnerabilities (the same as on any other WSGI server, of course).

PR #784, opened 2016-03-27
Test PyPy 5.0.1 on Travis

thedrow:
Just trying to keep the project up to date :)


jamadden:
Unfortunately it's not quite that simple because of the Travis caching (see https://travis-ci.org/gevent/gevent/jobs/118821188#L224). You'll need to change the directory name to get an absolutely reliable result. (I'd prefer not to blow away all the caches because it takes awhile to get them all back due to the way the caches race.)

PR #793, opened 2016-04-14
fix bug in PriorityQueue._init

Damnever:
When `items` is given, `PriorityQueue` can not maintain a correct heap.

```
import heapq
from gevent.queue import PriorityQueue

items = [(4, 'd'), (1, 'a'), (3, 'c'), (2, 'b')]

q = PriorityQueue(items=items)

print(q.get_nowait())

heapq.heapify(items)
print(heapq.heappop(items))
```

### DIFF
--- a/src/gevent/queue.py
+++ b/src/gevent/queue.py
@@ -375,6 +375,7 @@ class PriorityQueue(Queue):
     def _init(self, maxsize, items=None):
         if items:
             self.queue = list(items)
+            heapq.heapify(self.queue)
         else:
             self.queue = []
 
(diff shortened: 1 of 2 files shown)


jamadden:
Thanks. Can you add a test case?

src/gevent/queue.py
```diff
@@ -372,9 +372,11 @@ class PriorityQueue(Queue):
     Entries are typically tuples of the form: ``(priority number, data)``.
     '''
 
-    def _init(self, maxsize, items=None):
+    def _init(self, maxsize, items=None, heapify=heapq.heapify):
+        # pylint:disable=arguments-differ
```
jamadden: I know that the sibling methods `_put` and `_get` do the performance hack of copying a variable into a kwarg, and I appreciate your being consistent. 

In this case, though, I don't think it's warranted. Can you remove the kwarg and thus fix (instead of ignore) the warning?

jamadden:
Merged. Thank you!

PR #796, opened 2016-04-23
Consistently deal with LibreSSL in OpenSSL version tests

czarkoff:
Don't test OpenSSL version on LibreSSL systems: `ssl.OPENSSL_VERSION_INFO` on LibreSSL is always `(2, 0, 0, 0, 0)`, even for new versions then LibreSSL 2.0. Tested on OpenBSD -current with LibreSSL 2.4.0.

### DIFF
--- a/src/greentest/2.7.8/test_ssl.py
+++ b/src/greentest/2.7.8/test_ssl.py
@@ -189,8 +189,10 @@ class BasicSocketTests(unittest.TestCase):
         self.assertGreaterEqual(status, 0)
         self.assertLessEqual(status, 15)
         # Version string as returned by OpenSSL, the format might change
-        self.assertTrue(s.startswith("OpenSSL {:d}.{:d}.{:d}".format(major, minor, fix)),
-                        (s, t))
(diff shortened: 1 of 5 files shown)


jamadden:
Those files are exact copies of upstream CPython or PyPy as of the versions specified. Have these changes been made there? We try very hard not to modify these files because they get refreshed from upstream when new point releases are made (except for the frozen directory, 2.7.8).

jamadden:
Possibly a better approach would be a new section in `patched_test_setup.py` that removes this specific test case on platforms where it's not applicable.

czarkoff:
The issue is specific to LibreSSL, and it can be observed on all platforms supported by gevent. It may be worth considering disabling this test entirely - AFAICS it does not rely on gevent functionality in any way.

jamadden:
I would prefer not to disable it except when it is expected to fail, e.g., when LibreSSL is in use. The block in `patched_test_setup.py` ought to be able to take that into account, yes, if the test case itself can.

PR #797, opened 2016-04-24
Don't run 'test_openssl_version' when ssl is backed with LibreSSL

czarkoff:
On LibreSSL `OPENSSL_VERSION_INFO` is always `(2, 0, 0, 0, 0)`, so the test fails
for all LibreSSL versions after 2.0. In contrast to proper fix in #796, this PR works around the issue by disabling corresponding test in `src/greentest/patched_tests_setup.py` if `OPENSSL_VERSION.startswith('LibreSSL')`.


jamadden:
Thank you!

PR #798, opened 2016-04-25
Patch existing locks

lxsli:
In projects which dynamically determine whether to activate gevent,
it can be hard not to import a low level module like logging before
gevent. When logging is imported it initialises a threading.RLock
which it uses to protect the logging configuration. If two
greenthreads attempt to claim this lock, the second one will block the
/native/ thread not just itself. As green systems usually only have
one native thread, this will freeze the whole system.

While searching the GC for unsafe RLocks, replace their internal Lock
with a safe one. Or on Python3, where the default RLock implementation
is in C, swap the whole RLock out with a safe Python version.

### DIFF
--- a/src/gevent/monkey.py
+++ b/src/gevent/monkey.py
@@ -231,13 +231,17 @@ def patch_time():
 
 
 def _patch_existing_locks(threading):
+    import gevent.threading
     if len(list(threading.enumerate())) != 1:
         return
     try:
-        tid = threading.get_ident()
+        tid = gevent.threading.get_ident()
     except AttributeError:
-        tid = threading._get_ident()
+        tid = gevent.threading._get_ident()
+    lock_type = type(threading.Lock())
     rlock_type = type(threading.RLock())
+    if sys.version_info[0] >= 3:
+        pyrlock_type = type(threading._PyRLock())
     try:
         import importlib._bootstrap
     except ImportError:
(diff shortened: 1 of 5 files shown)


src/gevent/monkey.py
```diff
@@ -254,20 +258,53 @@ class _ModuleLock(object):
...
+    while old._is_owned():
+        old.release()
+        new.acquire()
+    if old._is_owned():
+        new.acquire()
+    gc.collect()
+    for ref in gc.get_referrers(old):
+        for k, v in vars(ref):
+            if v == old:
+                setattr(ref, k, new)
```
jamadden: This doesn't seem very robust. What if the ref is in a dict or list or set?

src/greentest/test__threading_holding_lock_while_monkey.py
```diff
@@ -1,8 +0,0 @@
-from gevent import monkey
```
jamadden: Why was this test removed?

src/gevent/monkey.py
```diff
@@ -231,13 +231,17 @@ def patch_time():
 
 
 def _patch_existing_locks(threading):
+    import gevent.threading
```
jamadden: This seems like a change that we don't need.

src/gevent/monkey.py
lxsli: Hmm, I could certainly look at extending this. The patch as-is provides an incremental improvement and (I hope) doesn't break anything.

src/greentest/test__threading_holding_lock_while_monkey.py
lxsli: My new test__monkey_oldlocks_locked.py provides more coverage than that test provided so I removed it to reduce redundancy.

jamadden:
Overall I'm not sure about this, especially the Py3 part. It seems like a lot of risky work (many ways it could subtlety fail) for a scenario that really isn't well supported to begin with (very-late monkey patching when threads are already in use). Maybe I'm missing something...

lxsli:
OK, let me give some background. I primarily work on OpenStack, which uses both eventlet and gevent in various projects. There's at least one project (Keystone) which decides whether to load eventlet based on its config. Loading config causes the logging module to be initialised, which causes an RLock to be created before monkey-patching. If a greenthread blocks on that RLock, the native thread locks which is disastrous.

So the tradeoff is some hairy code messing with GC traversal, to try and fix a lock which could freeze the whole system. If you follow the advice and monkey-patch first, the GC traversal won't find any RLocks so this patch is pretty much null-and-void.

Does that help?

lxsli:
BTW there should be no threads in use yet, only the single native thread. AFAIK spawning greenlets before monkey-patching is a terrible idea and continues to be.

jamadden:
> Loading config causes the logging module to be initialised, which causes an RLock to be created before monkey-patching.

But that particular lock is explicitly monkey-patched out already. Why is that not sufficient?

lxsli:
Ha, don't know how I missed that. Well - ok then! My patching is more comprehensive but for my purposes this is fine.

jamadden:
Cool! Thanks for the idea in any case.

PR #809, opened 2016-05-23
Invoke 'sh' instead of '/bin/sh' in deps setup

fornwall:
This makes building work on platforms lacking /bin/sh.

It specifically avoids problems when instaling gevent on [Termux](https://termux.com) where /bin/sh does not exist due to the Android file system layout. Problem reported in https://github.com/termux/termux-packages/issues/283.


jamadden:
Thanks! Merged with a162f35b1e9d8a2872fa63aa337d094f1faa36c8.

PR #814, opened 2016-05-29
Convert readthedocs link for their .org -> .io migration for hosted projects

adamchainz:
As per [their blog post of the 27th April](https://blog.readthedocs.com/securing-subdomains/) ‘Securing subdomains’:

> Starting today, Read the Docs will start hosting projects from subdomains on the domain readthedocs.io, instead of on readthedocs.org. This change addresses some security concerns around site cookies while hosting user generated data on the same domain as our dashboard.

Test Plan: Manually visited all the links I’ve modified.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -41,7 +41,7 @@ extensions = ['sphinx.ext.autodoc', 'sphinx.ext.doctest', 'sphinx.ext.coverage',
               'sphinx.ext.extlinks']
 
 intersphinx_mapping = {'http://docs.python.org/': None,
-                       'https://greenlet.readthedocs.org/en/latest/': None}
+                       'https://greenlet.readthedocs.io/en/latest/': None}
 
 extlinks = {'issue': ('https://github.com/gevent/gevent/issues/%s',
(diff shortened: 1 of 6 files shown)


jamadden:
Thank you!

I checked greenlet, cffi, and coverage on RTD.io, and all the HTTPS sites simply redirect (slowly) back to HTTP. Is there a documented plan that RTD.io will start serving over HTTPS? Otherwise I'd rather avoid the redirects.

adamchainz:
Weird, they don't redirect for me... maybe it's your browser? I'm on latest Chrome on Mac.

jamadden:
Hmm. So what I'm seeing is the redirect from the bare domain name (`greenlet.RTD.io`) to `/en/latest/`. It stays HTTPS, but I lose the padlock icon in Safari (it flashes then disappears)...but only for some domains! I could have sworn it was all the domains, but now I *keep* the padlock for greenlet, but lose it for cffi. Chrome shows the padlock for both domains, though. Maybe it's a browser extension or mixed content.

PR #826, opened 2016-07-08
Fix pypy3 imports

utzig:
In pypy3 import from `_thread` (python3) instead of `thread`.

### DIFF
--- a/src/gevent/lock.py
+++ b/src/gevent/lock.py
@@ -3,7 +3,7 @@
 from __future__ import absolute_import
 
 from gevent.hub import getcurrent
-from gevent._compat import PYPY
+from gevent._compat import PYPY, PY3
 from gevent._semaphore import Semaphore, BoundedSemaphore # pylint:disable=no-name-in-module,import-error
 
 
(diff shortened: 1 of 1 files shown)


jamadden:
Thank you for the PR and the interest in PyPy(3).

Sadly, we don't actually test on or support PyPy 3 at this time...the released versions of PyPy3 have the Python 3.2 stdlib, and the minimum supported version for gevent is Python 3.3.

I know that work is progressing for an updated PyPy3, and when one comes out that supports a more recent stdlib I'll gladly add it to the build matrix. But until then, this would just be dead code inviting bitrot, unfortunately.

jamadden:
Well heck, so I guess they've put out [an alpha](http://doc.pypy.org/en/latest/release-pypy3.3-v5.2-alpha1.html). I wonder if there are the portable linux images we need for it...

utzig:
You're correct, I'm using the alpha version which supports 3.3! https://morepypy.blogspot.com.br/2016/05/pypy33-v52-alpha-1-released.html

Anyway the patch is very simple and is working for me for both versions.

utzig:
I'm on a Mac. Did not test on Linux yet.

jamadden:
Were you able to run the complete test suite?

utzig:
I did not try but can do.

jamadden:
Yeah, that'd be the next step to get an idea of how complete it is for gevent's purposes:
```
cd src/greentest && GEVENT_RESOLVER=thread pypy3 testrunner.py --config known_failures.py
```

utzig:
https://gist.github.com/06be9e45781a279af74831384e74d93b

jamadden:
Thanks. So there's some work to do 😦 Some of those I remember from PyPy2 ('no such attribute timeout') but the actual hangs/timeouts aren't ringing any bells.

jamadden:
I myself do not currently have the time to follow up on these issues with PyPy3. I will personally consider them higher priority when a *beta* release comes out.

I would review a PR that fixed them and added a Travis test environment for it, but that's about the extent for me right now.

If some time opens up before the beta arrives I'll try to spend some time on this.

I'll leave this open as a reminder that it's not currently supported and work needs to be done.

PR #842, opened 2016-08-20
Only raise socket.gaierror from _socket3.connect_ex

### LINKED ISSUE #841: BlockingIOError raised by socket.connect_ex after monkey patch on python3.5
* gevent version: 1.1.2
* Python version: 3.5.0
* Operating System: mac os

### Description:

In non-blocking mode, `socket.connect_ex` is expected to return an errno and not raise an exception (except socket.gaierror caused by hostname / dns lookup failures).

On python3.5 gevent raises `BlockingIOError: [Errno 115] Operation now in progress` when it should simply return the errno, 115.

### What I've run:

```
from gevent import monkey
monkey.patch_all()

import socket
s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
s.setblocking(False)
ret = s.connect_ex(("www.google.com", 80))
print(ret)
```

It appears that gevent is attempting to return the error code here:
https://github.com/gevent/gevent/blob/b550598753d3cf4763cb47ba3bf804bfb9db26b0/src/gevent/_socket3.py#L305-L314 specifically, L312

But the `type(ex) is error` check is not sufficient. here, type(ex) is `BlockingIOError` while socket.error is `OSError`

dpkp:
Fix #841

### DIFF
--- a/src/gevent/_socket3.py
+++ b/src/gevent/_socket3.py
@@ -308,10 +308,10 @@ class socket(object):
         except timeout:
             return EAGAIN
         except error as ex:
-            if type(ex) is error: # pylint:disable=unidiomatic-typecheck
-                return ex.args[0]
-            else:
+            if type(ex) is gaierror: # pylint:disable=unidiomatic-typecheck
                 raise  # gaierror is not silented by connect_ex
+            else:
+                return ex.args[0]
 
     def recv(self, *args):
         while True:
--- a/src/greentest/test__socket.py
+++ b/src/greentest/test__socket.py
@@ -313,6 +313,20 @@ class TestFunctions(greentest.TestCase):
     # Creating new types in the function takes a cycle to cleanup.
     test_wait_timeout.ignore_leakcheck = True
 
+    def test_connect_ex(self):
+        # Issue 841
+        cli = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        ret = cli.connect_ex(('localhost', get_port()))
+        self.assertEqual(type(ret), int)
+        cli.close()
+
+    def test_connect_ex_gaierror(self):
+        # Issue 841
+        cli = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        with self.assertRaises(socket.gaierror):
+            cli.connect_ex(('foo.bar.fizzbuzz', get_port()))
+        cli.close()
+
 
 if __name__ == '__main__':
     greentest.main()


jamadden:
Please don't modify the files in greentest/2.x or greentest/3.x. Those are copies of the stdlib tests that we occasionally re-copy.

dpkp:
Ok -- tests failed as expected, which is all I really wanted to see. I'm happy to remove the tests if you like. Or is there a different place to write tests?

jamadden:
Please do revert those changes. gevent tests live in the files directly in greentest/, typically named with two underscores, as in `test__X`, e.g., `test__socket.py`

src/greentest/test__socket.py
```diff
@@ -313,6 +313,20 @@ class io(object):
     # Creating new types in the function takes a cycle to cleanup.
     test_wait_timeout.ignore_leakcheck = True
 
+    def test_connect_ex(self):
+        # Issue 841
+        cli = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self.addCleanup(cli.close)
```
jamadden: The test harness counts these sockets as leaked:
```
  ======================================================================
  FAIL: test_connect_ex (__main__.TestFunctions)
  ----------------------------------------------------------------------
    File "/home/travis/build/gevent/gevent/src/greentest/greentest.py", line 243, in wrapped
      raise AssertionError('refcount increased by %r\n%s' % (deltas, diff))
  AssertionError: refcount increased by [4, 4, 4]
  <type 'instancemethod'>: 62 != 63
  <type 'gevent.libev.corecext.io'>: 4 != 6
  <class 'gevent._socket2.socket'>: 2 != 3
  
  ======================================================================
  FAIL: test_connect_ex_gaierror (__main__.TestFunctions)
  ----------------------------------------------------------------------
    File "/home/travis/build/gevent/gevent/src/greentest/greentest.py", line 243, in wrapped
      raise AssertionError('refcount increased by %r\n%s' % (deltas, diff))
  AssertionError: refcount increased by [10, 436, 4, 4, 4]
  <type 'instancemethod'>: 65 != 66
  <type 'gevent.libev.corecext.io'>: 8 != 10
  <class 'gevent._socket2.socket'>: 4 != 5
```

You can use the gevent-specific function `self._close_on_teardown` to avoid this...it's also common practice in the gevent test suite to manually close the resource before the test function returns.

src/gevent/_socket3.py
```diff
@@ -308,10 +308,10 @@ def connect_ex(self, address):
         except timeout:
             return EAGAIN
         except error as ex:
-            if type(ex) is error: # pylint:disable=unidiomatic-typecheck
-                return ex.args[0]
-            else:
+            if type(ex) is gaierror: # pylint:disable=unidiomatic-typecheck
```
jamadden: Perusing cpython's `socketmodule.c`'s implementation of `connect_ex`, it appears that the address lookup is done in a separate step/function from the connect call. If *anything* goes wrong in the lookup step, an exception is (re)raised. If *anything* goes wrong in the connect step, the errno is returned.

Looking at address resolution, it appears that it can raise (at least) an OverflowError, a TypeError and possibly other errors that are address-family specific (there's a bunch of code to read there!). So if those errors are raised, they need to be re-raised by us. Here's an overflow error:
```python
>>> import socket
>>> s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
>>> s.setblocking(False)
>>> s.connect_ex(('www.google.com', 65789))
OverflowError: getsockaddrarg: port must be 0-65535.
```

So all of this is to say that the original code was closer to correct: return an errno value, but raise anything else. (With this patch as written, we'd wind up returning a str!) A corrected version *could* look something like:
```python
except error as ex:
  if isinstance(ex, error) and getattr(ex, 'errno', 0):
     return ex.errno
  raise
```

And of course it would need a corresponding test.

dpkp:
agree -- the internal semantics of connect_ex are more detailed than this naive fix. The underlying design problem is that gevent attempts to implement connect_ex as a wrapper around a patched connect. I think a more robust solution would be to refactor the patched connect_ex to call `_socket.socket.connect_ex` directly.

jamadden:
I disagree with that assessment. Cpython channels both connect and connect_ex through the same internal function, the only difference being a flag saying whether to raise or return an error.

dpkp:
Ok, I don't have strong opinions on how you go about fixing. I don't actually use gevent, so I'm probably not a great source of a PR to fix the problem!

jamadden:
Ok, no problem. Everybody has demands on their time. Thanks for the report and the effort you already put into a fix!

PR #848, opened 2016-08-30
lock: add an RWLock class for RW/RO locks.

ralt:
This lets an object have multiple kind of locks: read-write and
read-only. The idea is that an object can have multiple read-only locks,
but only a single read-write should be acquired. These locks are
mutually exclusive.

### DIFF
--- a/doc/intro.rst
+++ b/doc/intro.rst
@@ -211,10 +211,10 @@ I/O functions that bypass the libev event loop.
 Synchronizing access to objects shared across the greenlets is
 unnecessary in most cases (because yielding control is usually
 explict), thus traditional synchronization devices like the
-:class:`~lock.BoundedSemaphore`, :class:`~lock.RLock` and
-:class:`~lock.Semaphore` classes, although present, aren't used very
-often. Other abstractions from threading and multiprocessing remain
-useful in the cooperative world:
+:class:`~lock.BoundedSemaphore`, :class:`~lock.RLock`,
+:class:`~lock.RWLock`, and :class:`~lock.Semaphore` classes, although
+present, aren't used very often. Other abstractions from threading and
+multiprocessing remain useful in the cooperative world:
 
 - :class:`~event.Event` allows one to wake up a number of greenlets that are calling :meth:`~event.Event.wait` method.
 - :class:`~event.AsyncResult` is similar to :class:`~event.Event` but allows passing a value or an exception to the waiters.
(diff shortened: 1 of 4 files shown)


jamadden:
Please don't make modifications to the files in `greentest/2.7/` and `greentest/2.7pypy`, those files are copies of the tests shipped with the standard library so changes are likely to get lost when tests are updated.

ralt:
@jamadden thanks, that's fixed.

jamadden:
Do you have a concrete use case where a RW-lock improves concurrency above a simple lock in gevent's case? 

I've used [ReadWriteLocks](http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/ReadWriteLock.html) in Java before, and the conditions in which it offers a performance benefit are fairly narrow. Its [primary implementation](http://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/ReentrantReadWriteLock.html) goes into more detail about this, and also discusses many issues around how these locks are managed that would also need to be thought-through and probably documented here.

If we were to do this, instead of `acquire_ro` and `release_ro`, I like the approach that Java's ReadWriteLocks take, which is to expose a separate read lock and write lock. That way they can be used in `with` statements and even passed around separately as needed:

```python
rwlock = ...
with rwlock.read_lock():
   pass
call_method_that_locks(rwlock.write_lock())
```

ralt:
@jamadden the primary motivation is to introduce the concept of RO locks, as opposed to `RLock` which only lets you have a single lock at a time -- the classic case being reading a file. It's fine to have multiple read-only locks, but you want a read-write lock to lock all the other locks.

The public interface was chosen to duplicate what's used for `RLock`.

jamadden:
I believe I understand the motivation, but the main question is, is it actually worth it? Does it usefully increase concurrency in non-trivial use-cases? For example, reading a file doesn't provide an opportunity to switch greenlets, so it doesn't matter what kind of lock you use, it doesn't make a difference. I think this is a harder question to answer in a cooperatively multitasked system like gevent---and I *suspect* the answer is more likely to be "no."

I stand by my comments about the interface. As it is, this only *partially* matches `RLock` because it can't be used in a context manager (`with`). I think exposing two separate lock objects makes more sense than inventing a few new method names.

jamadden:
Closing this in triage for 1.2a1 since it hasn't been updated in two months. While the idea itself may have some merit, we need a more clear use case and better understanding of performance implications, plus the interface is under discussion.

PR #851, opened 2016-08-31
Don't inherit IMapUnordered from Greenlet

kerrick-lyft:
Instead, use the spawn method provided by the caller to spawn a manager greenlet. This ensures that the manager Greenlet is the same class that the caller uses.


jamadden:
To avoid surprises with incorrectly behaving user code, it is important that IMap continues to inherit from a regular Greenlet. I do not expect to accept this PR.

jamadden:
Thanks aga

PR #861, opened 2016-09-13
Encode c-ares/acountry.c in utf-8

jamadden:
Not in iso-8859 which is what it was identified as. There were
characters in a string constant that couldn't be decoded using utf-8. I
believe this will be a basically compatible change due to the nature of
the two encoding systems.

Workaround for https://github.com/landscapeio/dodgy/pull/7


jamadden:
That didn't solve the landscape issue. Instead, I guess the problem is with https://github.com/PyCQA/astroid/issues/361

PR #881, opened 2016-10-21
More tests for #879

sublee:
@jamadden Thank you for #879.  I read your work, then I thought we need more test cases for `gevent.os.waitpid`.  So I added such test cases for `waitpid`.

- `test_waitpid` is same with `test_waitpid_all` but it sets specific pid instead of `0`.
- `test_waitpid_timeout` tests cooperativity of `waitpid` using `gevent.Timeout`.

### DIFF
--- a/src/greentest/test__os.py
+++ b/src/greentest/test__os.py
@@ -97,6 +97,16 @@ if hasattr(os, 'fork_and_watch'):
                 gevent.sleep(2)
                 os._exit(0)
 
+        def test_waitpid(self):
+            pid = os.fork_and_watch()
+            if pid:
+                pid_, stat = os.waitpid(pid, 0)
+                assert pid_ == pid
+                os._reap_children(0)
+            else:
+                gevent.sleep(2)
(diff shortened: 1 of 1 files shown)


jamadden:
Thanks for your interest and test cases. Unfortunately, I will disagree with you: I don't feel like these are necessary.

- The specific pid case is extremely well covered by existing tests (see the coverage reports);
- I  *deliberately* did not add timeout tests or other tests to assert the cooperativity of specific cases, beyond what is needed to implement the higher-level guarantees of things like Popen. I consider that an implementation detail subject to change at any time; that's why the documentation is deliberately vague. So long as the code is covered, I'm happy.

FWIW, I also don't like that use of decorators to change method call signatures. There's nothing else in the gevent code base that does that. Were it necessary to de-duplicate the code (I don't really see that it is or would have been), a more idiomatic (simpler!) approach would be preferred.

sublee:
OK, I understand.  Thank you for reviewing it!

PR #884, opened 2016-10-24
Simplify subprocess.py: Unify many py2/py3 branches

jamadden:
Previous work added the context manager methods, which made this feasible. This adds a few extensions to the top-level API, notably `timeout` and `input` arguments. Otherwise it should be completely BWC as far as I can see.

Best viewed without whitespace, ?w=0.


jamadden:
There's one unrelated timing issue on Travis, otherwise we're green.

PR #896, opened 2016-11-28
Update libev to 4.23

jamadden:


jamadden:
The travis failures are due to timeouts in test_subprocess.py that I can't reproduce locally on OS X. They happen in master too.

PR #904, opened 2016-12-08
Add SSLContext to StreamServer et al

### LINKED ISSUE #901: Add SSLContext to StreamServer et al
* gevent version: 1.2
* Python version: 3.5+
* Operating System: Any

### Description:

Currently StreamServer and descendants take a whole bunch of SSL-related args but do not accept SSLConxext and gevent's wrap_socket does not accept it either even though both SSLSocket and SSLContext fully support both.

I'll be interested adding this myself if the commit will be accepted.

### What I've run:

I don't care what name it is, but I presume `context` would work as an addition to `**ssl_args` with appropriate changes here https://github.com/gevent/gevent/blob/master/src/gevent/server.py#L25 and further down the call stack.

arcivanov:
fixes #901

### DIFF
--- a/src/gevent/server.py
+++ b/src/gevent/server.py
@@ -20,8 +20,18 @@ class StreamServer(BaseServer):
     """A generic TCP server. Accepts connections on a listening socket and spawns user-provided *handle*
     for each connection with 2 arguments: the client socket and the client address.
 
-    If any of the following keyword arguments are present, then the server assumes SSL mode and uses these arguments
-    to create an SSL wrapper for the client socket before passing it to *handle*:
+    Server can assume an SSL mode via either direct client socket wrapper initialization or socket wrapping via an
+    :class:`SSLContext`.
+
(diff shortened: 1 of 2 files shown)


jamadden:
Thanks! That's a lot simpler than I was imagining.

There are some build issues on Travis that will go away if you rebase on current master.

There are some test failures both [on Travis](https://travis-ci.org/gevent/gevent/jobs/182195292#L5874) and [AppVeyor (windows)](https://ci.appveyor.com/project/denik/gevent/build/1.0.820/job/66ktfq1op8515m2q#L6071) that need to be addressed.

Python 2.7.8 doesn't have `ssl.create_default_context`, so [those tests](https://github.com/gevent/gevent/pull/904/files#diff-d73d714891cc4d7b9e09cc2307c0b809R747) need to be conditionally defined only when that's present.

jamadden:
I have a branch where these things should be fixed, I think. I'll finish testing and merge from there.

PR #910, opened 2016-12-13
Release the GIL on DNS operations

thedrow:


jamadden:
The ares operations are asynchronous (non-blocking) already, aren't they? (That's sort of the point.) So I really don't see much benefit to this, unless the setup they do is *very* heavyweight (I don't think they are, but I don't know). It seems to me it has the potential to hurt performance by adding extra contesting over the GIL.

I'd prefer to see some proof that this has a measurable, positive difference before merging something that changes the GIL semantics.

thedrow:
It seems that Cython prevents us from releasing the GIL anyway. Unless I'm doing something that is unreasonable.

With a CFFI extension the GIL would have been released anyway. So maybe we should match the behaviour?

jamadden:
Ah, but there is no CFFI extension, so there's nothing to match 😄

thedrow:
I'll get to it. Promise :)

jamadden:
I've just started looking at it, but I'm liking the idea of a dnspython-based resolver. 

- Being pure-python, it can't crash the interpreter.
- Being pure-python, it has no lengthy build and configure step like c-ares.
- Being pure-python, it's automatically portable to PyPy/CPython/IronPython/Jython.
- It's almost certainly simpler than the ares integration, which has to do interesting tricks to get the event loops to cooperate.
- Likewise, it's *probably* faster than the ares integration (the ares integration is much slower than the threadpool-based resolver).

thedrow:
Really? That's a surprise. c-ares is used everywhere.

thedrow:
It's not possible to acquire the GIL in the callback with cython.
I'm +1 for dnspython.

jamadden:
It turns out that #1088 is *much* faster than c-ares on all interpreters. Although it doesn't yet handle host files, that may slow it down some.

PR #913, opened 2016-12-15
timeout remaining

mayflaver:
Add remaining method to Timeout


jamadden:
Thank you for the PR! I'm disinclined to accept this unless there is a very strong use case for the simple reason that libuv, an alternate event loop implementation that we're exploring, has no equivalent of `ev_timer_remaining`.

It also seems somewhat less than useful in general because it applies to the loop's notion of the current time, which is only updated occasionally.

jamadden:
Thanks again for this PR. The reasons stated in the first comment haven't changed, and there's been no other discussion, so I'm going to go ahead and close it now.

PR #939, opened 2017-02-14
fix(subprocess): make sure empty value for PY3 and text mode is correct

william-gr:
Without this patch, if a subprocess is in text mode and there is no stderr or stdout output it will default to empty bytes (b"") in python 3, instead of empty string ("").

### DIFF
--- a/src/gevent/subprocess.py
+++ b/src/gevent/subprocess.py
@@ -702,10 +702,12 @@ class Popen(object):
         self._stdout_buffer = None
         stderr_value = self._stderr_buffer
         self._stderr_buffer = None
-        # XXX: Under python 3 in universal newlines mode we should be
-        # returning str, not bytes
-        return (None if stdout is None else stdout_value or b'',
-                None if stderr is None else stderr_value or b'')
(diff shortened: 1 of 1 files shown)


jamadden:
It seems to me that there is a reason for this behavior. Can you provide a test that passes on unpatched Python 2 and 3, but fails with a patched 2 and 3?

william-gr:
Hm, not sure what you're asking...

Take this for instance:
```
proc = subprocess.PIPE("ls 1>&2", shell=True, stdout=subprocess.PIPE, stderr=subprocess.PIPE, encoding='utf-8')
proc.communicate()[0].split('\n')
```

Without this patch it will fail on python 3, because stdout will be b'', even though text mode was specified (encoding='utf8'). With this patch it will work because it will return "", as expected.
This is a no-op for python 2.

william-gr:
@jamadden Oh, one more thing I forgot. This mimics the real python 3 subprocess behavior for empty stdout/stderr. Not sure if my fix is the better one since I am not sharp on gevent internals, but thats what I would expect as a end result.

PR #958, opened 2017-03-26
Bump PyPy version to 5.7 for both PyPy2 and PyPy3

thedrow:
Note that there is a regression discovered with PyPy 5.7 that will be fixed soon.
Maybe we should hold off merging this until the patch version is released.


thedrow:
I'm not sure how you installed the Python versions since the dependencies to compile it are missing.
Also, I'm not sure what it's trying to compile since the default PyPy versions are the portable versions.
Any idea?

jamadden:
Locally I'm getting crashes in the test suite running with PyPy2 5.7.1. This will need some more looking into (installation issues aside).

PR #959, opened 2017-03-26
Enable ccache to shorten build times

thedrow:
I was sure I contributed it but I didn't. It should be an easy win.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ matrix:
   fast_finish: true
 
 script:
-    - make $TASK
+    - make $TASK CC='ccache gcc'
 
 notifications:
   email: false
@@ -40,3 +40,4 @@ cache:
 
 before_cache:
     - rm -f $HOME/.cache/pip/log/debug.log
+    - ccache -s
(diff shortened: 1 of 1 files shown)


jamadden:
Without also forcing make and the like to use the `ccache` binary instead of `gcc`, I don't think this is going to do anything. For some parts of some builds it may be enough to `export CC=ccache`, but it's not clear to me that that would do handle everything.

thedrow:
It seems like it didn't. I'll make sure it does.

thedrow:
@jamadden I don't think the build failure is related.

jamadden:
It appears that in the best cases, this might possibly save maybe 20 seconds or so. But there's so much variability in our build times already that it's hard to be sure.

It was definitely worth a try (thanks!), but since the win is so small, I'd be inclined to not merge this: I'd prefer not to introduce more differences between local development and the CI environments than we have to, and I'm not going to try to bother with installing ccache locally 😄

thedrow:
ccache doesn't affect anything other than compilation times. I don't think it's a big deal that your development environment doesn't have it.
I'll close the PR for now. If you reconsider, reopen and merge it.

PR #967, opened 2017-04-12
Ignore exceptions in `ThreadPool._worker`

damz:
When the runtime tears down, it is somewhat likely that a
ThreadPool._worker will trigger an exception. When the runtime
tries to output that exception to sys.stderr, it is also
somewhat likely that this output will fail, with an error
similar to:

```
Unhandled exception in thread started by
sys.excepthook is missing
lost sys.stderr
```

This can currently be reproduced with something as simple as:

```
while true; do python -c 'import gevent.hub ; gevent.hub.get_hub().threadpool.apply(lambda: True)' done
```

### DIFF
--- a/src/gevent/threadpool.py
+++ b/src/gevent/threadpool.py
@@ -192,47 +192,53 @@ class ThreadPool(GroupMappingMixin):
     _destroy_worker_hub = False
 
     def _worker(self):
-        # pylint:disable=too-many-branches
-        need_decrease = True
         try:
-            while True:
-                task_queue = self.task_queue
-                task = task_queue.get()
-                try:
-                    if task is None:
-                        need_decrease = False
-                        self._decrease_size()
-                        # we want first to decrease size, then decrease unfinished_tasks
-                        # otherwise, _adjust might think there's one more idle thread that
-                        # needs to be killed
-                        return
-                    func, args, kwargs, thread_result = task
+            # pylint:disable=too-many-branches
+            need_decrease = True
+            try:
+                while True:
+                    task_queue = self.task_queue
+                    task = task_queue.get()
                     try:
-                        value = func(*args, **kwargs)
-                    except: # pylint:disable=bare-except
-                        exc_info = getattr(sys, 'exc_info', None)
-                        if exc_info is None:
+                        if task is None:
+                            need_decrease = False
+                            self._decrease_size()
+                            # we want first to decrease size, then decrease unfinished_tasks
+                            # otherwise, _adjust might think there's one more idle thread that
+                            # needs to be killed
                             return
-                        thread_result.handle_error((self, func), exc_info())
-                    else:
-                        if sys is None:
-                            return
-                        thread_result.set(value)
-                        del value
+                        func, args, kwargs, thread_result = task
+                        try:
+                            value = func(*args, **kwargs)
+                        except: # pylint:disable=bare-except
+                            exc_info = getattr(sys, 'exc_info', None)
+                            if exc_info is None:
+                                return
+                            thread_result.handle_error((self, func), exc_info())
+                        else:
+                            if sys is None:
+                                return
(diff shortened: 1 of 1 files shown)


jamadden:
Thanks for the report! 

I can't reproduce the behaviour with the example you provide using any of Python 2.7.13, 3.4.6, or 3.5.3 (on OS X) after several minutes of running. Can you provide some more details about the environment where you're seeing this?

damz:
@jamadden We are seeing this under Python 2.7 on a variety of Debian-based Linux environments. It is a race between the main thread tearing down the environment and the threads for the threadpool, so it is possible that OS X and Linux are taking different scheduling decisions.

If you have a working Docker installation, you should be able to reproduce with this piece of stupidity:

```
docker run --rm debian:jessie bash -c 'set -e ; apt-get update ; export DEBIAN_FRONTEND=noninteractive ; apt-get -y install python python-pip python-dev ; pip install gevent ; while true; do python -c "import gevent.hub ; gevent.hub.get_hub().threadpool.apply(lambda: True)" ; done'
```

(then kill it by getting the ID of the container with `docker ps`, followed by `docker rm -f <id>`.)

damz:
I just tried on python-3.4 on that same Debian Jessie container, and it doesn't show the symptom. Not sure if the issue is still present there, or if they just removed the infamous `lost sys.stderr` message printed directly to stderr.

jamadden:
Does it effect the process exit status, or is it just printed? 

Can you reproduce it in Python 3?

jamadden:
I was also able to reproduce this on Python 2.7.9 on a Linux machine, at a rate of about 1 in 15 times. I verified that the exit code was unaffected.

I was not able to reproduce it on Python 3.4 or above. I believe this is because those versions support [PEP 442](https://docs.python.org/3/whatsnew/3.4.html#pep-442-safe-object-finalization) which does not set module globals to `None` (I did confirm that the same [two](https://github.com/python/cpython/blob/3.4/Python/pythonrun.c#L1870) [messages](https://github.com/python/cpython/blob/3.4/Python/pythonrun.c#L2066) can still theoretically be printed on 3.4 and 3.5).

I totally sympathize with the desire for no unwanted error output (and I may have a solution for you, see below!), but all of that points to what I see as some fairly high bars that this change would need to overcome to be part of the core:

1. This adds another layer of nesting to an already complex function;
2. The problem appears solved for newer versions of Python (so eventually this would become legacy code for no apparent reason);
3. Testing this does what it is intended would be complex, if possible (which is bad given 2);
4. It appears to be fairly rare, and doesn't impact the exit status when it does happen

So here's hopefully the good news, a possible alternative solution. The Hub allows you to configure the ThreadPool class it uses, either by setting the environment variable `GEVENT_THREADPOOL` to the FQN of a subclass of ThreadPool, or, if you do so early enough, by setting `get_hub().threadpool_class` to the FQN. So it should be possible to do something like this:

```python
class QuietThreadPool(ThreadPool):
   def _worker(self):
      try:
         ThreadPool._worker(self)
      except:
          pass
```

``` $ GEVENT_THREADPOOL=module.QuietThreadPool python -c "import gevent.hub ; gevent.hub.get_hub().threadpool.apply(lambda: True)"```

Hopefully that can work for you.

damz:
@jamadden Setting the module globals to `None` is definitely one of the races we have seen. It doesn't seem to be the most common one. I haven't been able to precisely track the others down.

Anyway, this is fine. We do a large number of additional monkey patching on top of `gevent`'s (for `open`, `os.fsync` and `os.fdopen`), so adding a bit more will not hurt us here.

jamadden:
The thing about the module globals not being `None` means that if there is an uncaught exception, `sys.excepthook` should still be around (as should `sys.stderr`) so it would be reported with a reasonable traceback. Hopefully.

PR #969, opened 2017-05-11
FileObjectClosed is an exception object

ralt:
So, raise it directly, instead of trying to use it as a callable.


jamadden:
Thank you! The lint failures on Travis are unrelated.

PR #975, opened 2017-05-25
monkey: exec scripts in global scope

sbohrer:
Commit 59581a8a3b7b0a4722fb3f7074fce6cd4c700bfd "Add python 3.5
selectors test, and remove the devpoll selector under 3.5. also update
the monkey documentation." Introduced a regression by moving the exec of
scripts into a main() function. If the script being run with exec has
import statements those are now locally scoped to the main() function
which can break the script.

It does not appear that this local scoping was intentional so the fix is
to exec the script in the global scope, returning the previous behavior.

### DIFF
--- a/src/gevent/monkey.py
+++ b/src/gevent/monkey.py
@@ -672,8 +672,10 @@ def main():
         __package__ = None
         assert __package__ is None
         globals()['__file__'] = sys.argv[0]  # issue #302
+        # Clear package so exec'd script doesn't appear in gevent package
+        globals()['__package__'] = None
         with open(sys.argv[0]) as f:
-            exec(f.read())
+            exec(f.read(), globals())
     else:
         print(script_help)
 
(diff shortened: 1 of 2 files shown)


sbohrer:
Well, this breaks test__issue302monkey.py because it causes __package__ == gevent  and I'm not exactly sure what the correct fix is.  I could always do this:

globals()['__package__'] = None

before the exec, but I have no idea if that is the correct thing to do.  Strangely going back to gevent 1.0 which does not have the problem I am trying to fix because there was no main() function, invoking the exec there with the globals() does not cause __package__ to become gevent.

jamadden:
I don't know off the top of my head what the correct thing is either, All I can suggest right now is trying your proposal.

gevent 1.0 is different than 1.1+ because it didn't support Python 3 and used `execfile`, a function that no longer exists IIRC.

jamadden:
Looks like that works! Can you add a change note and a line comment in monkey.py  about `__package__` (like the one for `__file__`)?

sbohrer:
I had already updated the commit description, but I added a comment to the code.

jamadden:
Thanks!

PR #979, opened 2017-06-07
always obey GEVENT_NO_CFFI_BUILD

danc86:
With `LIBEV_EMBED=0`, `GEVENT_NO_CFFI_BUILD=1`, and the `cffi` module absent, every invocation of setup.py *still* tries to configure the embedded copy of libev. Looks like the test for `IGNORE_CFFI` is just in the wrong place, should be outside the try: else: block. This patch moves it around.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -90,16 +90,17 @@ try:
 except ImportError:
     pass
 else:
-    if IGNORE_CFFI and not PYPY:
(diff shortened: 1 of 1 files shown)


jamadden:
LGTM (test failures unrelated). Thanks!

PR #999, opened 2017-07-21
Suppress [Errno 10053] errors (Windows)

jan11011977:
This prevents pywsgi from printing stack traces when the client disconnects from the server.

When running pywsgi on Windows I am seeing a lot of these stack traces:
```
File "c:\virtualenv\compass\2017-07-20T00.23.18\lib\site-packages\gevent\pywsgi.py", line 935, in handle_one_response
    self.run_application()
  File "c:\virtualenv\compass\2017-07-20T00.23.18\lib\site-packages\geventwebsocket\handler.py", line 87, in run_application
    return super(WebSocketHandler, self).run_application()
  File "c:\virtualenv\compass\2017-07-20T00.23.18\lib\site-packages\gevent\pywsgi.py", line 909, in run_application
    self.process_result()
  File "c:\virtualenv\compass\2017-07-20T00.23.18\lib\site-packages\gevent\pywsgi.py", line 895, in process_result
    self.write(data)
  File "c:\virtualenv\compass\2017-07-20T00.23.18\lib\site-packages\gevent\pywsgi.py", line 741, in write
    self._write_with_headers(data)
  File "c:\virtualenv\compass\2017-07-20T00.23.18\lib\site-packages\gevent\pywsgi.py", line 763, in _write_with_headers
    self._write(data)
  File "c:\virtualenv\compass\2017-07-20T00.23.18\lib\site-packages\gevent\pywsgi.py", line 726, in _write
    self._sendall(data)
  File "c:\virtualenv\compass\2017-07-20T00.23.18\lib\site-packages\gevent\pywsgi.py", line 701, in _sendall
    self.socket.sendall(data)
  File "c:\virtualenv\compass\2017-07-20T00.23.18\lib\site-packages\gevent\_socket2.py", line 419, in sendall
    timeleft = self.__send_chunk(chunk, flags, timeleft, end)
  File "c:\virtualenv\compass\2017-07-20T00.23.18\lib\site-packages\gevent\_socket2.py", line 355, in __send_chunk
    data_sent += self.send(chunk, flags)
  File "c:\virtualenv\compass\2017-07-20T00.23.18\lib\site-packages\gevent\_socket2.py", line 323, in send
    return sock.send(data, flags)
error: [Errno 10053] An established connection was aborted by the software in your host machine
Thu Jul 20 11:02:00 2017 {'REMOTE_PORT': '58259', 'HTTP_HOST': 'shg-compass', 'REMOTE_ADDR': '10.222.13.18', (hidden keys: 33)} failed with error
```
I can see that we are already suppressing EPIPE and ECONNRESET errors, but on Windows there is also the WSAECONNABORTED which needs to be suppressed.

Thanks!

### DIFF
--- a/src/gevent/pywsgi.py
+++ b/src/gevent/pywsgi.py
@@ -946,8 +946,8 @@ class WSGIHandler(object):
         except _InvalidClientInput:
             self._send_error_response_if_possible(400)
         except socket.error as ex:
-            if ex.args[0] in (errno.EPIPE, errno.ECONNRESET):
-                # Broken pipe, connection reset by peer.
+            if ex.args[0] in (errno.EPIPE, errno.ECONNRESET, errno.WSAECONNABORTED):
+                # Broken pipe, connection reset by peer, connection aborted.
                 # Swallow these silently to avoid spewing
                 # useless info on normal operating conditions,
                 # bloating logfiles. See https://github.com/gevent/gevent/pull/377


jamadden:
The reason that we feel good about swallowing EPIPE and ECONNRESET is that they're usually caused by the remote machine hanging up, i.e., the client timed out and quit, or lost its connection, or whatever. But the error message here ("connection was aborted by the software in your host machine") seems to suggest that it's the local machine/process that's at fault, which sounds very different and more concerning to me---I would want to see if there was something I needed to adjust locally to stop those errors.

Then there's the fact that the `WSA` stuff isn't defined outside of Windows, hence all the Travis test failures. Checking for it would need to be made conditional. One way to do it would be with `getattr(errno, 'WSAwhatever', errno.EPIPE)`, but that's ugly and perpetuates hard-coding of this growing list.

Instead we could use a value on self:
```python
class Server(object):
    swallowed_connection_abort_errnos = (errno.EPIPE, errno.ECONNRESET)
    if hasattr(errno, 'WSAwhatever'):
       swallowed_connection_abort_ernos += (errno.WSAwhatever,)
    ...
    def ...:
        if ex.args[0] in self.swallowed_connection_abort_errnos:
             # swallow
```

I like that approach for a few reasons. One is that it gives us a place to use a doc-comment so we can document what we're doing and why. Another is that we could actually leave out WSAwhetever (if it truly is of a different kind than the other caught exceptions) and allow users to subclass or set an instance variable if they did need to catch it.

I'm happy to do the first part for sure (extract to a ivar and document).

jan11011977:
Great! Thank you for your comments

Your suggestion of extracting the list of exceptions into a variable makes complete sense. I didn't realise that WSA* was a Windows-only exception although it seems obvious now that you mention it.

The Windows "_connection was aborted by the software in your host machine_" error is rather badly worded. It sounds like something has gone wrong with the local machine, but it actually means that the remote client closed the connection while our server was writing to it. 

Some further explanation is here:
https://www.chilkatsoft.com/p/p_299.asp

PR #1000, opened 2017-07-21
Extract list of ignored exceptions in pywsgi to an ivar. Include errno.WSAECONNABORTED on Windows.

jamadden:
Fixes #999.


jan11011977:
Awesome. Many thanks.

PR #1029, opened 2017-10-17
Test with PyPy 5.9

thedrow:
This PR bumps our tested PyPy version to 5.9 for both Python 2 & 3 versions.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,34 +165,34 @@ develop:
 	${PIP} install -U -r dev-requirements.txt
 
 lint-py27: $(PY27)
-	PYTHON=python2.7.13 PATH=$(BUILD_RUNTIMES)/versions/python2.7.13/bin:$(PATH) make develop travis_test_linters
+	PYTHON=python2.7.13 PATH=$PY27/bin:$(PATH) make develop travis_test_linters
 
 test-py27: $(PY27)
-	PYTHON=python2.7.14 PATH=$(BUILD_RUNTIMES)/versions/python2.7.14/bin:$(PATH) make develop fulltoxtest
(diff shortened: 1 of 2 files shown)


thedrow:
Seems like there are problems with PyPy 5.9 to fix.

jamadden:
Unfortunately, this didn't actually run anything with PyPy 5.9. [It still ran with 5.8](https://travis-ci.org/gevent/gevent/jobs/288949463#L598). That's because the Makefile also needs to be adjusted to the new path. (Why use distinct paths? Caching. If we don't, we have to explicitly clear the build caches to change a version.)

I'm not too concerned about the failures, a re-run would probably clear them up as they look to be timing related. However, it might be time to [update the test suite](https://travis-ci.org/gevent/gevent/jobs/288949463#L619).

thedrow:
Well it still fails in ways I wouldn't expect it to fail. It now fails to import gevent :(

jamadden:
The problem now is that [the wrong pip](https://github.com/python/cpython/blob/master/Python/ceval.c#L4140) is being used to install:
```
echo pip is at `which pip`
pip is at /home/travis/.runtimes/versions/pypy3.5_580/bin/pip
echo python is at `which /home/travis/.runtimes/snakepit/pypy3.5_590`
python is at /home/travis/.runtimes/snakepit/pypy3.5_590
```
We're getting the pip from 5.8, but using the Python from 5.9.

There's one more spot in the Makefile that needs to be changed to specify the correct PATH for both test-pypy and test-pypy3 targets.

thedrow:
It is very likely that I am just too tired to commit without mistakes.
I'll try tomorrow. Maybe I'll revert the change to generalize the paths because it doesn't work (TM).

PR #1033, opened 2017-11-09
Fix EPROTOTYPE race condition on macOS

parasyte:
I have an integration test that runs a very simple HTTP server with `gevent.pywsgi` and [bottle](https://bottlepy.org). During one of my test runs (and only one, so far) this exception occurred on macOS 10.12.6:

```
  File "/Users/jay/.virtualenvs/venv/lib/python2.7/site-packages/gevent/pywsgi.py", line 935, in handle_one_response
    self.run_application()
  File "/Users/jay/.virtualenvs/venv/lib/python2.7/site-packages/gevent/pywsgi.py", line 909, in run_application
    self.process_result()
  File "/Users/jay/.virtualenvs/venv/lib/python2.7/site-packages/gevent/pywsgi.py", line 895, in process_result
    self.write(data)
  File "/Users/jay/.virtualenvs/venv/lib/python2.7/site-packages/gevent/pywsgi.py", line 741, in write
    self._write_with_headers(data)
  File "/Users/jay/.virtualenvs/venv/lib/python2.7/site-packages/gevent/pywsgi.py", line 763, in _write_with_headers
    self._write(data)
  File "/Users/jay/.virtualenvs/venv/lib/python2.7/site-packages/gevent/pywsgi.py", line 726, in _write
    self._sendall(data)
  File "/Users/jay/.virtualenvs/venv/lib/python2.7/site-packages/gevent/pywsgi.py", line 701, in _sendall
    self.socket.sendall(data)
  File "/Users/jay/.virtualenvs/venv/lib/python2.7/site-packages/gevent/_socket2.py", line 419, in sendall
    timeleft = self.__send_chunk(chunk, flags, timeleft, end)
  File "/Users/jay/.virtualenvs/venv/lib/python2.7/site-packages/gevent/_socket2.py", line 355, in __send_chunk
    data_sent += self.send(chunk, flags)
  File "/Users/jay/.virtualenvs/venv/lib/python2.7/site-packages/gevent/_socket2.py", line 323, in send
    return sock.send(data, flags)
error: [Errno 41] Protocol wrong type for socket
Thu Nov  9 11:50:12 2017 {'REMOTE_PORT': '54471', 'HTTP_HOST': '127.0.0.1:8000', 'REMOTE_ADDR': '::ffff:127.0.0.1', (hidden keys: 23)} failed with error
```

According to [this blog post](http://erickt.github.io/blog/2014/11/19/adventures-in-debugging-a-potential-osx-kernel-bug/), `EPROTOTYPE` is an expected errno from the `write` syscall on macOS! It occurs on a very specific race condition, which my integration tests are capable of triggering.

Digging into gevent source, it will correctly handle `EPIPE` and `ECONNRESET` in 1.2.2: https://github.com/gevent/gevent/blob/1.2.2/src/gevent/pywsgi.py#L949 and 1.3 (currently unreleased) adds `WSAECONNABORTED` for Windows: https://github.com/gevent/gevent/blob/e913b3f64b353dcf1c2bbff1d58762ae81a4ed4b/src/gevent/pywsgi.py#L924-L941

This patch will catch `EPROTOTYPE` and retry the write, which should raise `EPIPE` as expected. Not included in the patch is a test which demonstrates the bug. I would have committed the tests directly to `greentest`, but I just couldn't get `make test-py27` to work.

This test can repro the issue; it runs for about a minute (on my laptop) when the patch is applied, then exits with a success message. It typically fails within 15 seconds without the patch.

```python
from gevent import monkey
monkey.patch_all()

import errno
import socket
import threading

addr = ("", 8000)

def child():
    c = socket.create_connection(addr)
    c.close()

def main():
    s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
    s.bind(addr)
    s.listen(128)

    threading.Thread(target=child).start()

    (conn, _) = s.accept()
    while True:
        try:
            conn.send(b"\0")
        except socket.error as e:
            assert e.args[0] == errno.EPIPE, "error: %s" % (e)
            break

    conn.close()
    s.close()

for _ in range(25):
    main()

print("Done, no errors!")
```

Further information: The same issue was [reported to nodeJS](https://github.com/nodejs/node/issues/2382) a few years ago, and they patched `libuv` to handle it: https://github.com/libuv/libuv/blob/8a9585232e9df227bcaf1263748d8997ec5d1077/src/unix/stream.c#L829-L836

### DIFF
--- a/src/gevent/_socket2.py
+++ b/src/gevent/_socket2.py
@@ -5,6 +5,7 @@ Python 2 socket module.
 # Our import magic sadly makes this warning useless
 # pylint: disable=undefined-variable
 
+import errno
 import time
 from gevent import _socketcommon
 from gevent._util import copy_globals
@@ -322,7 +323,9 @@ class socket(object):
         try:
             return sock.send(data, flags)
         except error as ex:
-            if ex.args[0] != EWOULDBLOCK or timeout == 0.0:
+            # macOS can return EPROTOTYPE when writing to a socket that is shutting
+            # Down. Retrying the write should return the expected EPIPE error.
+            if ex.args[0] not in (EWOULDBLOCK, errno.EPROTOTYPE) or timeout == 0.0:
                 raise
             sys.exc_clear()
             self._wait(self._write_event)
--- a/src/gevent/_socket3.py
+++ b/src/gevent/_socket3.py
@@ -7,6 +7,7 @@ Python 3 socket module.
 # pylint: disable=too-many-statements,too-many-branches
 # pylint: disable=too-many-public-methods,unused-argument
 from __future__ import absolute_import
+import errno
 import io
 import os
 import sys
@@ -389,7 +390,9 @@ class socket(object):
         try:
             return _socket.socket.send(self._sock, data, flags)
         except error as ex:
-            if ex.args[0] != EWOULDBLOCK or timeout == 0.0:
+            # macOS can return EPROTOTYPE when writing to a socket that is shutting
+            # Down. Retrying the write should return the expected EPIPE error.
+            if ex.args[0] not in (EWOULDBLOCK, errno.EPROTOTYPE) or timeout == 0.0:
                 raise
             self._wait(self._write_event)
             try:


jamadden:
I can (sadly) confirm that the issue still exists in macOS 10.13.1.

Since this is a mac-specific issue, it seems better to me that the check for the additional errno only happens on mac (much like libuv did). Perhaps that could look like this:
```python
# _socketcommon.py
GESENDAGAIN = (EWOULDBLOCK,)
if sys.platform == 'darwin':
      # macOS can return EPROTOTYPE when writing to a socket that is shutting
      # Down. Retrying the write should return the expected EPIPE error.
  from errno import EPROTOTYPE
  GESENDAGAIN += (EPROTOTYPE,)

# socket2 and socket3.py
def send(...):
  try:
     ...
  except IOError as e:
    if e.args[0] not in _socketcommon.GSENDAGAIN or timeout == 0.0:
      raise
    ...
```

If you rebase on master, Travis CI tests should pass.

parasyte:
@jamadden No problem. I wasn't certain that introducing `sys.platform` checks was the best solution, since I didn't see it used anywhere else other than tests. On the plus side, I agree the patch should be macOS-specific. And thanks for the tip about `_socketcommon`, seems like a good place to front-load the conditionals.

I'm happy to rebase against master, but Travis is still reporting some failures. When it all goes green, I'll create a new PR and close this one out!

jamadden:
> I wasn't certain that introducing sys.platform checks was the best solution, since I didn't see it used anywhere else other than tests.

Yeah, I can't say I really *like* it, but it does seem best to keep this limited, and that's the only way.

> I'm happy to rebase against master, but Travis is still reporting some failures.

Feel free to go ahead and rebase. The failures are intermittent and load-related. They tend to clear up depending on the time of day and if tests environments get run individually.

parasyte:
Out of curiosity, did you just force-push the last commit on master? I had to rewrite that commit on my fork's master branch (and force-push) because the commit hash changed.

Whatever the case, I'm wrapping up the rebased patched now.

PR #1038, opened 2017-11-14
Group.add: ignore completed greenlets

RonRothman:
This PR modified `Group.add(greenlet)` (hence, transitively, `Pool.add()`) to ignore any greenlet that has already completed.

[This discussion](https://github.com/gevent/gevent/pull/1032#issuecomment-343490241) got me thinking that we might benefit from checking this condition. Let me know what you think. Feel free to merge, close, or suggest improvements as you see fit. :)


jamadden:
Thanks for the PR.

This would result in subtly changed semantics for what seems a very minor benefit so I'm not sure it's worth merging (plus it has a bug; see below). Currently, adding a completed greenlet to a Group will show the greenlet as a member of the group until the event loop gains control. (The act of calling `rawlink` schedules a callback in the event loop.) This would break that. In general, completed greenlets can show as members of the pool at any time, so this doesn't eliminate needing to deal with that case.

Because this is checked only in Group.add, not Pool.add, we could still block and switch greenlets waiting on the pool's semaphore, so it doesn't seem to assist with concurrency management. 

Sadly, it also introduces a bug in the Pool: We've already acquired the semaphore by the time we get here, and the only way to release the semaphore is to add it to the pool so its callback is executed. This short-circuits that, so we have a case where we might acquire the entire semaphore and not actually have any greenlets in the pool.

RonRothman:
Yikes, this is true. I was thinking of the semaphore as a mere mutex, not as a counting device that depends (via callback) upon a greenlet actually being added.

I'd consider moving the new check into `Pool.add`, before we acquire the semaphore, but I agree with you, it's not worth the minor benefit. Will close. Thanks for looking!

Issue #790, opened 2016-04-10
libuv backend


jamadden:
In theory, libuv [provides a better windows experience](https://github.com/gevent/gevent/issues/580#issuecomment-203386688) and [may be faster even on unix](https://github.com/joyent/libuv/issues/485#issuecomment-6717239).

jamadden:
(copied from https://github.com/gevent/gevent/issues/580#issuecomment-207112223):

I've been looking at libuv, and it's not quite the panacea one would hope. 

First, what's relevant for this thread, the DNS support, sadly, is also no better: it just shoves requests off onto its own internal threadpool. Making it no different than gevent's default resolver.

Other minor problems include no support for fork watchers or child watchers; those can both be handled at the gevent level though, at least on POSIX (but the situation with children and SIGCHLD won't get any better).

The biggest impedance mismatch, though, is in the handling of IO. The libuv equivalent to libev's IO watchers, "poll" watchers, pretty much matches exactly what libev does, including only working for sockets on Windows (and not using IOCP, AFAICS). If you want the much-vaunted Windows improvements you have to use functions and types specific to TCP, UDP or pipes, and these *do not* have (exposed) read or write events, relying instead on a series of non-blocking callbacks that are handled internally to the event loop. This doesn't fit well with gevent's "wait for a read event then try to read again" approach (in fact, gevent wouldn't be able to use the Python-level socket object *at all*). It may be possible to paper over this with some higher-level abstractions, but they don't currently exist.

So in the short term, I'm still interested in looking into a libuv port based on the poll watchers. I think that's pretty low hanging fruit, and I'm interested to compare performance. For another thing, this should make CFFI possible on Windows. But any other Windows-specific improvements that would come from redesigning the socket abstractions are further down the line.

jamadden:
I've hit a wall with libuv: you can't use libuv in a forked child process, period, no workaround, end-of-story, the child process crashes. See https://github.com/gevent/gevent/commit/550b698e4f3b517c5950fcf8ba6235384f84f7b6#diff-89fa73d9850232e80e8a3ef604f73eaeR106 .

[...]

PR #1048, opened 2017-12-05
Initial libuv support

jamadden:
This is still very experimental; see the change note for some details and #790 for some other details.

I am inclined to merge this to master and work on further refinements in other PRs. Since it is opt-in that probably shouldn't cause too many problems. There are changes here that I want in master, such as the refactoring of the loops and watchers, and there are things that I want to do in master for libev (such as changing to new-style callbacks) that would definitely introduce conflicts.


jamadden:
We have green builds on appveyor and travis!

PR #1050, opened 2017-12-08
Use setuptools to build libuv, and enable it on Windows (especially for PyPy)

jamadden:
This is the current approach used by https://github.com/saghul/pyuv/, and it worked locally for me with CPython 3.5 on Windows 10---I was *never*able to get the `vcbuild.bat` version to work for either CPython or PyPy locally or on appveyor. CPython always had linking errors and PyPy never got to the point of building the project.

I've only tested CPython locally, but I would expect it to also work for PyPy---we'll see if appveyor ever gets around to building us. This is exciting because it's our first support for PyPy on Windows.

The downside is that it slightly increases our maintenance burden when updating libuv sources---we have to reconfirm the included files, macros and library settings; however, it finally works, so I'm ok with that for now. It also removes any ability to use a system libuv, at least for the time being. That can probably be added back eventually if someone cares, but statically linking everything in and using `-flto` seems to produce a nice performance boost.


jamadden:
Well it builds and runs, but most of the socket tests fail with `ENOTSOCK`. We're using `uv_poll_init`, which calls `_get_osfhandle()` to transform the `socket.fileno()` result back into a `SOCKET` handle, so something is presumably going wrong during that process.

I will note that libev, when not embedded in gevent, does something similar on Win32 (though whether it works or not I don't know, I think the tests only run in embedded mode).

jamadden:
Current status: PyPy gets mysterious socket errors, as mentioned in previous commits and comments. PyPy is only 32-bit on windows right now.

Testing Python 3.6 both 32 and 64-bit with libuv on test_ftplib.py *mostly* works (much more so that pypy does), only having a few (~8) timeout errors. Interestingly, there are some stack traces from the SSL tests in test_ftplib.close (404) `super(SSLConnection, self).close()` line 356 goes to asyncore.py line 397 `self.socket.close()` which generates an AttributeError: 'int' object has no attribute 'close'. So that's fishy.

But we don't actually test test_ftplib.py under libev on Windows right now, so it may have the same issue. Lets find out.

jamadden:
(Moved from an EDIT to the previous comment to a new comment, in case anyone is following via email.)

EDIT: Yes, libev does have the same issues. There are 7 timeouts instead of 8, and there are the same SSLConnection.close -> int.close() issues. Plus there's this, which is odd:
```
test_ccc (__main__.TestTLS_FTPClass) ... Exception in thread Thread-83:
  File "e:\githubsources\gevent-main\src\gevent\select.py", line 92, in _make_watchers
    self.__add_watchers(watchers, rlist, self.add_read, io, MAXPRI)
  File "e:\githubsources\gevent-main\src\gevent\select.py", line 81, in __add_watchers
    watcher = io(get_fileno(fd), callback.event)
  File "gevent.libev.corecext.pyx", line 488, in gevent.libev.corecext.loop.io
  File "gevent.libev.corecext.pyx", line 836, in gevent.libev.corecext.io.__init__
OSError: 736 is not a socket (files are not supported)
```

So pretty much libuv on CPython and libev on CPython are equivalent. PyPy is still misbehaving. I'm getting close to commiting this (but disabling/ignoring tests) and moving on, probably back to CPython.

PR #1056, opened 2017-12-15
Support libuv under CPython 3.6 on Windows

jamadden:
This required some changes to the way the IO watcher manages its memory to avoid GC cycles and ensure timely cleanups so that libuv `uv_poll` handles don't get reused for different SOCKET objects with the same fileno(). 

I disabled PyPy on Windows for now since it takes so long to run (and fail), once I realized that PyPy didn't actually work on Posix either, probably for these kind of determinism issues. That's going to take more work to get PyPy working correctly since we can't count on the determinism, but we're going in the right direction. (The ultimate solution will probably look like how `socket.makefile` and `SocketIO` cooperate on Python 3.)

This in turn exposed some issues in error handling and signals under libuv (but *potentially* also under libev).

There is still an occasional failure in libuv that I have seen on all platforms (windows 3.6, posix 2.7 and 3.6):
```
  ERROR: test_exception (__main__.TestCallback)
  ----------------------------------------------------------------------
    File "test__systemerror.py", line 76, in tearDown
      assert not self.x.pending, self.x
  AssertionError: <callback at 0x2b5c6b880390 pending callback=<function raise_ at 0x2b5c6e7fc668> args=(Exception('regular exception must not kill the program',),)>
  
  ----------------------------------------------------------------------
```

pylint is failing on travis because 1.8 slipped out. I'll fix that in  the next PR.

### DIFF
--- a/examples/concurrent_download.py
+++ b/examples/concurrent_download.py
@@ -11,7 +11,13 @@ monkey.patch_all()
 
 import sys
 
-urls = ['http://www.google.com', 'http://www.yandex.ru', 'http://www.python.org']
+# Note that all of these redirect to HTTPS, so
+# this demonstrates that SSL works.
+urls = [
+    'http://www.google.com',
+    'http://www.apple.com',
+    'http://www.python.org'
+]
 
 
 if sys.version_info[0] == 3:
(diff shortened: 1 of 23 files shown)


examples/concurrent_download.py
```diff
@@ -11,7 +11,13 @@
 
 import sys
 
-urls = ['http://www.google.com', 'http://www.yandex.ru', 'http://www.python.org']
```
jamadden: yandex.ru was removed because windows couldn't load its certificate.

PR #1081, opened 2018-01-26
Try to adhere to scheduling deadlines when running callbacks

### LINKED ISSUE #1068: time.sleep(0.000001, False) => 'This operation would block forever'
* gevent version: 1.2.2
* Python version: PyEnv Python 3.6.4 (default, Dec 19 2017, 17:09:19) [GCC 7.2.1 20170915 (Red Hat 7.2.1-2)] on linux
* Operating System: Fedora 27  4.14.11-300.fc27.x86_64 #1 SMP Wed Jan 3 13:52:28 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

### Description:

Given: a multi-threaded application that spawns couple of servers (nginx and nghttpx) via Popen with stdout and stderr piped, spawns threads to read said pipes and flush them into asynchronous Fluent logger (which spawns another thread that feeds from a queue of log messages being sent to it). 

The entire application is monkey-patched first thing in `__main__` as follows so all threads are greenlets:

```
if __name__ == '__main__':
    from gevent import monkey
    monkey.patch_all(Event=True)

    import time

    _sleep = time.sleep

    def sleep(seconds):
        return _sleep(seconds, False)

    time.sleep = sleep
```  

The application on `m4.large`  (2 CPUs) seems to fail to send messages to FluentD (it's been verified they are entering the Fluent thread queue) and on `r4.2xlarge` (8 CPUs) the messages are delayed. 

One theory I have is that the FluentD logger thread is starved by other threads. So I added a `thread.sleep(0.000001)` in the main thread after FluentD initialization to allow FluentD thread to be scheduled.

Alas, what I got was:

```
File "/usr/local/lib/python3.6/site-packages/<snip>/manager.py", line 526, in init_logging
time.sleep(0.000001) # Gevent
File "/<main.py>", line 14, in sleep
return _sleep(seconds, False)
File "/usr/local/lib/python3.6/site-packages/gevent/hub.py", line 169, in sleep
hub.wait(loop.timer(seconds, ref=ref))
File "/usr/local/lib/python3.6/site-packages/gevent/hub.py", line 651, in wait
result = waiter.get()
File "/usr/local/lib/python3.6/site-packages/gevent/hub.py", line 898, in get
return self.hub.switch()
File "/usr/local/lib/python3.6/site-packages/gevent/hub.py", line 630, in switch
return RawGreenlet.switch(self)
gevent.hub.LoopExit: ('This operation would block forever', <Hub at 0x7f68a973df20 epoll default pending=0 ref=0 fileno=3 resolver=<gevent.resolver_thread.Resolver at 0x7f68a6fab0f0 pool=<ThreadPool at 0x7f68a6fab160 0/1/10>> threadpool=<ThreadPool at 0x7f68a6fab160 0/1/10>>)
```

Under what circumstances **ever** would a 1 usec sleep block forever?! 
Did I forget to patch something?

### LINKED ISSUE #1071: greenlet using time.sleep seems starving
* gevent version: 1.2.2
* Python version: CPython, 2.7.14, Ubuntu 17.10 package
* Operating System: Ubuntu 17.10 (Linux 4.13.0-21-generic #24-Ubuntu SMP Mon Dec 18 17:29:16 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux)

### Description:

Consider this example: it is the smallest possible example which reproduce an issue which I found on a real system; at its core, there are two greenlets:

1. `reader`: it `recv()` a chunk of bytes from a socket, then simulates a CPU-intensive processing. Every ~0.0005 seconds, it `time.sleep(0)` to yield control to other greenlets. Every ~1 second, it does a `recv()`.

2. `heartbeat`: endless loop which prints a message and does `time.sleep(0.1)`

```python
import sys
import time
import socket

HUB = sys.argv[1]
assert HUB in ('eventlet', 'gevent')

if HUB == 'eventlet':
	import eventlet
	from eventlet import spawn
	eventlet.sleep()
	eventlet.monkey_patch()
else:
	import gevent.monkey
	from gevent import spawn
	gevent.monkey.patch_all()


def reader():
	chunksize = 1024 * 100
	conn = socket.create_connection(('localhost', 3997))
	while True:
		print
		print 'Reading from socket...'
		chunk = conn.recv(chunksize)
		# simulate a CPU-intensive processing of incoming data
		for i in range(2000):
			busy_loop(0.0005)
			time.sleep(0)


def busy_loop(t):
	start = time.time()
	end = start + t
	while time.time() < end:
		pass


def heartbeat():
	while True:
		a = time.time()
		time.sleep(0.1)
		b = time.time()
		print 'heartbeat [waited for %.4f secs]' % (b-a)


def main():
	greenthread = spawn(reader)
	spawn(heartbeat)
	if HUB == 'gevent':
		greenthread.join()
	else:
		greenthread.wait()

if __name__ == '__main__':
	main()
```

### What I've run:

To simulate a fast server, I ran `nc` on localhost:

```
$ cat /dev/urandom | nc -l 3997
```

I would expect to see approximately 10 `heartbeat` for every `Reading from socket...`.
If I run it with `eventlet`, it seems to work as I expect:

```
$ python starve.py eventlet

Reading from socket...
heartbeat [waited for 0.1000 secs]
heartbeat [waited for 0.1001 secs]
heartbeat [waited for 0.1003 secs]
heartbeat [waited for 0.1004 secs]
heartbeat [waited for 0.1000 secs]
heartbeat [waited for 0.1003 secs]
heartbeat [waited for 0.1002 secs]
heartbeat [waited for 0.1003 secs]
heartbeat [waited for 0.1002 secs]
heartbeat [waited for 0.1001 secs]

Reading from socket...
heartbeat [waited for 0.1002 secs]
heartbeat [waited for 0.1002 secs]
heartbeat [waited for 0.1002 secs]
heartbeat [waited for 0.1001 secs]
heartbeat [waited for 0.1003 secs]
heartbeat [waited for 0.1001 secs]
heartbeat [waited for 0.1002 secs]
heartbeat [waited for 0.1005 secs]
heartbeat [waited for 0.1000 secs]
heartbeat [waited for 0.1004 secs]
```

However, if I try it with `gevent`, I see this weird behavior:
```
$ python starve.py gevent

Reading from socket...
heartbeat [waited for 0.5114 secs]

Reading from socket...
heartbeat [waited for 0.5094 secs]
heartbeat [waited for 0.5089 secs]

Reading from socket...
heartbeat [waited for 0.5087 secs]
heartbeat [waited for 0.5086 secs]

Reading from socket...
heartbeat [waited for 0.5085 secs]
heartbeat [waited for 0.5085 secs]

Reading from socket...
heartbeat [waited for 0.5087 secs]
heartbeat [waited for 0.5084 secs]
```

So, `heartbeat` is scheduled much less frequently than it should. I don't know if this is the intended behavior or a bug, but it certainly looks weird to me.

### LINKED ISSUE #1072: Can/Should `gevent.sleep` be smarter?
See #1071 for a concise explanation of the current behaviour (with reference to #1068 for some more discussion). It allows the programmer to tune the system, but it can sometimes be surprising. There are also changes coming with libuv where it enforces a 1ms minimum sleep time that make this a more important issue to address: old idioms like `time.sleep(0.00000001)` to cycle the event loop will not work as expected under libuv.

One possibility is to have a hybrid approach to the number of callbacks we run per event loop iteration, periodically checking for how long we've run and cycling the loop if some threshold is exceeded. Presumably that threshold, and really the count, should be configurable.

libev offers a way to set a minimum blocking time in its select call that may sometimes also be useful (though not directly related to this). Not sure if libuv has anything similar.

jamadden:
This should result in the loop getting serviced more often when there are many callbacks to run. This is hooked up to `getswitchinterval` to allow control.

We check the interval every 50 callbacks. That's hardcoded, which I don't like, but the 1000 callback number was hardcoded before, so it's not really much of a change. A future extension would be to calculate the check number dynamically based on dividing the switch interval.

I don't have a specific test case for this change because I'm a little fried (burnt out), but all the other tests pass. I did manually verify that the example in #1071 matches expectations---that could probably be turned into a test case with some work.

Fixes #1072
Fixes #1071
Fixes #1068

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -160,6 +160,16 @@ Other Changes
 - gevent now uses cffi's "extern 'Python'" callbacks. These should be
   faster and more stable. This requires at least cffi 1.4.0. See :issue:`1049`.
 
+- gevent now approximately tries to stick to a scheduling interval
+  when running callbacks, instead of simply running a count of
+  callbacks. The interval is determined by
+  :func:`gevent.getswitchinterval`. On Python 3, this is the same as
+  the thread switch interval. On Python 2, this defaults to 0.005s and
+  can be changed with :func:`gevent.setswitchinterval`. This should
+  result in more fair "scheduling" of greenlets, especially when
+  ``gevent.sleep(0)`` or other busy callbacks are in use. The interval
+  is checked every 50 callbacks to keep overhead low. See
+  :issue:`1072`. With thanks to Arcadiy Ivanov and Antonio Cuni.
 
 libuv
 -----
@@ -238,6 +248,8 @@ libuv
     time updated, without cycling the event loop. See :issue:`1057`.
 
     libev has also been changed to follow this behaviour.
+
+    Also see :issue:`1072`.
 
   - Timers of zero duration do not necessarily cause the event loop to
     cycle, as they do in libev. Instead, they may be called
--- a/src/gevent/__init__.py
+++ b/src/gevent/__init__.py
@@ -22,24 +22,28 @@ version_info = _version_info(1, 3, 0, 'dev', 0)
 __version__ = '1.3.0.dev0'
 
 
-__all__ = ['get_hub',
-           'Greenlet',
-           'GreenletExit',
-           'spawn',
-           'spawn_later',
-           'spawn_raw',
-           'iwait',
-           'wait',
-           'killall',
-           'Timeout',
-           'with_timeout',
-           'getcurrent',
-           'sleep',
-           'idle',
-           'kill',
-           'signal',
-           'fork',
-           'reinit']
+__all__ = [
+    'get_hub',
+    'Greenlet',
+    'GreenletExit',
+    'spawn',
+    'spawn_later',
+    'spawn_raw',
+    'iwait',
+    'wait',
+    'killall',
+    'Timeout',
+    'with_timeout',
+    'getcurrent',
+    'sleep',
+    'idle',
+    'kill',
+    'signal',
+    'fork',
+    'reinit',
+    'getswitchinterval',
+    'setswitchinterval',
+]
 
 
 import sys
@@ -47,6 +51,26 @@ if sys.platform == 'win32':
     # trigger WSAStartup call
     import socket  # pylint:disable=unused-import,useless-suppression
     del socket
+
+try:
+    # Floating point number, in number of seconds,
+    # like time.time
+    getswitchinterval = sys.getswitchinterval
+    setswitchinterval = sys.setswitchinterval
+except AttributeError:
+    # Running on Python 2
+    _switchinterval = 0.005
+
+    def getswitchinterval():
+        return _switchinterval
+
+    def setswitchinterval(interval):
+        # Weed out None and non-numbers. This is not
+        # exactly exception compatible with the Python 3
+        # versions.
+        if interval > 0:
+            global _switchinterval
+            _switchinterval = interval
 
 from gevent.hub import get_hub, iwait, wait
 from gevent.greenlet import Greenlet, joinall, killall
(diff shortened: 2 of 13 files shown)


jamadden:
Should there be no feedback (and should I be up early enough—appveyor builds are most reliable in my AM hours) I intend to tag and release 1.3a1 tomorrow.

PR #1088, opened 2018-02-01
Add a DNSpython resolver

### LINKED ISSUE #580: CFFI bindings for c-ares

jamadden:
This adds a resolver implementation using dnspython.org.

The hope was that it would be simpler than c-ares (it is, especially once https://github.com/rthalley/dnspython/pull/300 lands, if it does) and also faster than the threaded resolver.

Here's a trivial benchmark, lifted from the test case:
```python

def resolve(res):
    for index in range(100):
        try:
            res.gethostbyname('www.x%s.com' % index)
        except socket.error:
            pass
```

All times in seconds, smallest is best, hitting Google's public DNS server (8.8.8.8):

| Resolver  | PyPy 2 5.10 | CPython 2.7.14 |
|-----------|-------------:|----------------:|
| Blocking  |       74.76 |          61.04 |
| c-ares    |       50.17 |          53.26 |
| thread    |      158.41 |         122.06 |
| dnspython |      161.80 |         155.97 |

So, dnspython is not near as fast as the system (blocking) resolver or the c-ares resolver (even on pypy), but it has overhead comparable to the threaded resolver.

dnspython uses a cache of results. If I change the benchmark to just use a range of 20 and call it twice in a row, I get these results:

| Resolver  | PyPy 2 5.10 | CPython 2.7.14 |
|-----------|-------------:|----------------:|
| Blocking  |       0.0565 |        0.0586   |
| c-ares    |       1.7725 |          6.4246 |
| thread    |      0.1165 |         0.1707 |
| dnspython |      1.1844 |         9.2374 |

On PyPy, faster than c-ares now. Note that these results are *wildly* variable and extremely suspect. For example, on one run I had ares at 8.4s and dnspython at 1.4, but the next run had ares at 2.2s and dnspython at 23.6s. See below for more reliable numbers.

dnspython is *just* a DNS library, it does not use `/etc/hosts` at all. I could see that being a desirable trait for reproducibility. It currently is reading `/etc/resolv.conf` , but we may be able to add environment variables to configure it.  Another thing to note is that it only works in monkey-patched processes.

This needs some further documentation updates before merging, and I'm not 100% convinced that it's really worth merging (the refactoring, yes, the dnspython, maybe not). Feedback is welcome.

Fixes #580.


arcivanov:
Is a loop of 100 iterations enough to make PyPy finish jitting and stabilize?

jamadden:
> Is a loop of 100 iterations enough to make PyPy finish jitting and stabilize?

Excellent question. I think the answer is "no", based on this vmprof trace (where I'm actually running 200 iterations): http://vmprof.com/#/556030b6-df67-4e71-bc38-120621e3cfbe

But there was so much noise from the DNS servers that it was basically pointless to measure anyway. I've set up a caching DNS server locally and am working to get some more reliable numbers.

jamadden:
New numbers! [This script](https://gist.github.com/jamadden/b369cd5c5679969190481ef9ba7fd1b2) is hitting a single DNS server, a local dnsmasq instance configured to have everything cached. The result is numbers that are much more stable. The default backends were used (corecext and corecffi).

First, PyPy ([vmprof trace](http://vmprof.com/#/6271a48a-0152-4e71-a529-55c6d4cecbc1)):

| Resolver | 1st 100 resolves | 300 resolves |  Last 100 resolves|
| -------- | ------------: | ---------------: | --------------: |
| dnspython |      0.29 |      0.81 |      0.18 |
|  blocking |      0.11 |      0.31 |      0.10 |
|      ares |      1.14 |      3.58 |      1.21 |
|    thread |      0.13 |      0.35 |      0.12 |

Now CPython 2.7:

| Resolver | 1st 100 resolves | 300 resolves |  Last 100 resolves|
| -------- | ------------: | ---------------: | --------------: |
| dnspython |      0.39 |      1.80 |      0.51 |
|  blocking |      0.12 |      0.31 |      0.11 |
|      ares |      1.12 |      3.60 |      0.98 |
|    thread |      0.13 |      0.35 |      0.12 |

Finally, CPython 3.7b1:

| Resolver | 1st 100 resolves | 300 resolves |  Last 100 resolves|
| -------- | ------------: | ---------------: | --------------: |
| dnspython |      0.37 |      1.08 |      0.23 |
|  blocking |      0.12 |      0.31 |      0.11 |
|      ares |      1.11 |      3.42 |      0.96 |
|    thread |      0.13 |      0.35 |      0.12 |

The middle column is probably the best column to focus on, it's the total of 300 sequential `gethostbyname` calls. All three platforms have the same story: the native blocking resolver is the fastest (on macOS, there's a layer of caching in the name resolution library itself), followed closely by the threaded resolver (the extra overhead comes from synchronization), followed by dnspython, lagging by 2-3x, followed *far* behind by c-ares.

So if the threaded resolver doesn't work for you, and you can live without access to `/etc/hosts` (or you can run a dnsmasq instance to provide that), the dnspython option might be a good one for you.

Now, the above benchmarks are completely sequential, not demonstrating any parallelism at all. What happens if we batch up 100 greenlets and 100 DNS queries at the same time? 

PyPy (I had the ares resolver crash on me at least once in this scenario):

| Resolver | 1st 100 resolves | 300 resolves |  Last 100 resolves|
| -------- | ------------: | ---------------: | --------------: |
| dnspython |      0.05 |      0.18 |      0.06 |
|  blocking |      0.11 |      0.31 |      0.11 |
|      ares |      0.80 |      2.33 |      0.77 |
|    thread |      0.04 |      0.13 |      0.05 |

CPython 2.7

| Resolver | 1st 100 resolves | 300 resolves |  Last 100 resolves|
| -------- | ------------: | ---------------: | --------------: |
| dnspython |      0.08 |      0.29 |      0.09 |
|  blocking |      0.12 |      0.32 |      0.11 |
|      ares |      0.85 |      2.49 |      0.83 |
|    thread |      0.04 |      0.11 |      0.04 |

CPython 3.7

| Resolver | 1st 100 resolves | 300 resolves |  Last 100 resolves|
| -------- | ------------: | ---------------: | --------------: |
| dnspython |      0.07 |      0.21 |      0.06 |
|  blocking |      0.12 |      0.32 |      0.11 |
|      ares |      0.81 |      2.45 |      0.84 |
|    thread |      0.04 |      0.11 |      0.04 |

To put that in relative terms, how much faster was the parallel operation? (What percent of the sequential time did parallel take)?

| Resolver | PyPy | 2.7 |  3.7 |
| -------- | ------------: | ---------------: | --------------: |
| dnspython |      22% |      16% |      19% |
|  blocking |      100% |      100% |      100% |
|      ares |      65% |      69% |      71% |
|    thread |      37% |      31% |      31% |

So the dnspython resolver appears to scale better than the ares resolver or the threaded resolver.

arcivanov:
DNSPython provides another valuable feature - TTL-obeying cache and the one you can reset. System caches are often not quite compliant and impossible to expire. This is a great addition IMO.

Could you also add latest production CPython (3.6) to bench?

jamadden:
> DNSPython provides another valuable feature - TTL-obeying cache and the one you can reset. System caches are often not quite compliant and impossible to expire.

A salient point.

> Could you also add latest production CPython (3.6) to bench? 

OK.

> 3.7 is expected to have an additional 20% performance gain due to new method-resolution instruction set which would not be the case in 3.6.

I know 😉

jamadden:
> (Note that under PyPy if I let an exception bubble up, the ares process could crash.)

Reported at https://bitbucket.org/pypy/pypy/issues/2745/cpyext-exceptions-in-greenlets-running

jamadden:
If I enable the dnspython cache, PyPy's time goes from 0.18 to 0.07 for parallel, while 2.7 goes from 0.29 to 0.07 and 3.6 goes from 0.21 to 0.05.

Sequential time effectively becomes the same as the parallel time, after the first iteration.

The dnspython cache appears to respect the TTL of the DNS answers; although it does incur a syscall (`time.time`) for each access.

jamadden:
eventlet [has code to read hosts files](https://github.com/eventlet/eventlet/blob/master/eventlet/support/greendns.py). An ambitious next step would be for someone to port that to gevent.

carsonip:
@jamadden I decided to try to reproduce the results you've got here. I had issues with thread dns resolver as mentioned [here](http://carsonip.me/posts/gevent-built-in-function-getaddrinfo-failed-with-error/) then I moved to ares. In my experience, ares is fast and stable. I wanted to check if dnspython can do better than ares.

However, dnspython performs horribly on my machine. Here's the result using your benchmarking script:
```
Testing dnspython
    18.34    5.46    7.78    6.78    17.17
dnspython: best of 5 runs: 5.46; worst: 18.34
Testing blocking
    1.10    0.34    0.40    0.63    0.41
blocking: best of 5 runs: 0.34; worst: 1.10
Testing ares
    0.33    0.55    0.31    0.33    0.42
ares: best of 5 runs: 0.31; worst: 0.55
Testing thread
    0.43    0.40    0.42    0.45    0.47
thread: best of 5 runs: 0.40; worst: 0.47
| Resolver | One Iteration | Three Iterations | Delta 3 - two |
| -------- | ------------: | ---------------: | --------------: |
| dnspython |      2.36 |      5.46 |      1.61 |
|  blocking |      0.01 |      0.34 |      0.32 |
|      ares |      0.01 |      0.31 |      0.29 |
|    thread |      0.03 |      0.40 |      0.34 |
```
dnspython is at least 10x slower than all other resolvers. I tried with dnspython 1.15.0 and also github master branch.

Setup:
Linux Mint 19 (based on Ubuntu 18.04) with default system dns cache at 127.0.0.53
Python 2.7.15rc1
gevent==1.3.6

Any clue? Thanks.

jamadden:
The earlier script evolved into a repeatable benchmark [shipped with gevent sources](https://github.com/gevent/gevent/blob/master/benchmarks/bench_dns_resolver.py). I run it against a local dnsmasq server configured to cache everything (I found there was a *tremendous* degree of variability based on caches):
```bash
$ sudo dnsmasq -d --cache-size=100000 --local-ttl=1000000 --neg-ttl=10000000 --max-ttl=100000000 --min-cache-ttl=10000000000  --no-poll --auth-ttl=100000000000
```

And then I run the benchmark script like so: (you may want to tweak the values of the two constants to determine how long you want the script to run; don't forget to edit `/etc/resolv.conf` for the blocking and threaded resolvers):
```bash
$ GEVENT_RESOLVER_NAMESERVERS=127.0.0.1 python benchmarks/bench_dns_resolver.py --inherit-environ GEVENT_RESOLVER_NAMESERVERS
```

Running in Python 3.7.0 on my macOS machine  this morning with `N=150` and `RUN_COUNT=2`, I get (results grouped and ordered):
```
dnspython sequential: Mean +- std dev: 147 us +- 4 us
blocking sequential : Mean +- std dev: 1.11 ms +- 0.03 ms
thread sequential   : Mean +- std dev: 1.21 ms +- 0.02 ms
ares sequential     : Mean +- std dev: 6.56 ms +- 0.21 ms

dnspython parallel  : Mean +- std dev: 152 us +- 7 us
thread parallel     : Mean +- std dev: 357 us +- 16 us
blocking parallel   : Mean +- std dev: 1.11 ms +- 0.01 ms
ares parallel       : Mean +- std dev: 6.76 ms +- 0.21 ms
```

carsonip:
I found myself interested in the benchmark again. This time I ran the script shipped with gevent sources. `N=150`, `RUN_COUNT=2`. I still find ares super fast and consistent on my linux machine.
```
......
dnspython sequential: Mean +- std dev: 100 us +- 4 us
......
dnspython parallel: Mean +- std dev: 103 us +- 2 us
......
WARNING: the benchmark result may be unstable
* the standard deviation (33.5 ms) is 32% of the mean (104 ms)
* the maximum (211 ms) is 103% greater than the mean (104 ms)

Try to rerun the benchmark with more runs, values and/or loops.
Run 'python -m perf system tune' command to reduce the system jitter.
Use perf stats, perf dump and perf hist to analyze results.
Use --quiet option to hide these warnings.

blocking sequential: Mean +- std dev: 104 ms +- 33 ms
......
WARNING: the benchmark result may be unstable
* the standard deviation (23.9 ms) is 26% of the mean (92.0 ms)
* the minimum (40.1 ms) is 56% smaller than the mean (92.0 ms)
* the maximum (156 ms) is 70% greater than the mean (92.0 ms)

Try to rerun the benchmark with more runs, values and/or loops.
Run 'python -m perf system tune' command to reduce the system jitter.
Use perf stats, perf dump and perf hist to analyze results.
Use --quiet option to hide these warnings.

blocking parallel: Mean +- std dev: 92.0 ms +- 23.9 ms
......
ares sequential: Mean +- std dev: 36.7 us +- 2.0 us
......
ares parallel: Mean +- std dev: 32.8 us +- 1.8 us
......
WARNING: the benchmark result may be unstable
* the standard deviation (11.9 ms) is 369% of the mean (3.23 ms)
* the minimum (128 us) is 96% smaller than the mean (3.23 ms)
* the maximum (46.3 ms) is 1333% greater than the mean (3.23 ms)

Try to rerun the benchmark with more runs, values and/or loops.
Run 'python -m perf system tune' command to reduce the system jitter.
Use perf stats, perf dump and perf hist to analyze results.
Use --quiet option to hide these warnings.

thread sequential: Mean +- std dev: 3.23 ms +- 11.91 ms
......
thread parallel: Mean +- std dev: 79.4 us +- 7.5 us
```

Linux Mint 19.1 (based on Ubuntu 18.04) with dnsmasq started with the options specified in the script
Python 2.7.15rc1
gevent==1.3.7

PR #1096, opened 2018-02-12
Centralize configuration in a ``gevent.config`` object.

### LINKED ISSUE #1090: FEATURE: Allow API-based control of gevent-wide behaviors without env vars
Quite a few features of Gevent rely on setting env vars, such as DNS resolver, thread pool, etc.
This gets very cumbersome when monkeypatch bootstrap is used, pollutes process's own `os.environ` with GEVENT settings that would be inherited by children by default, which is not always desirable.

Ideally, gevent would have two methods of configuring global behavioral settings that compliment each other and allow configuring from API.

Possible example (not advocating any particular module/naming convention):
```
from gevent import settings

current_setting = settings.get_gevent_xxx()
settings.set_gevent_xxx(classname_or_str_or_value)

from gevent.monkey import ...
```

The `get_gevent_xxx` would first check if the setting was set via API, then via environ, then assume default.

Another advantage is that settings being aggregated that way in one place are easier to document. Right now it's almost impossible to be sure you found documentation on all of the configurable behaviors of gevent - have to grep repo code by hand.

jamadden:
This lets us document things more clearly, and, with care, allows configuring gevent from Python code without any environment varibles.

All the fileobject implementations needed to be moved out of the ``gevent.fileobject`` module so as not to introduce import cycles that made configuring from code impossible.

Fixes #1090

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -46,6 +46,18 @@
   :mod:`time`, much like :mod:`gevent.socket` can be imported instead
   of :mod:`socket`.
 
+- Centralize all gevent configuration in an object at
+  ``gevent.config``, allowing for gevent to be configured through code
+  and not *necessarily* environment variables, and also provide a
+  centralized place for documentation. See :issue:`1090`.
+
+  - The new ``GEVENT_CORE_CFFI_ONLY`` environment variable has been
+    replaced with the pre-existing ``GEVENT_LOOP`` environment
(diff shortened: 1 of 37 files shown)


arcivanov:
You sir are a god among us mortals!

jamadden:
Thank you for the suggestion! I think it makes gevent better.

PR #1102, opened 2018-02-15
Make socket.sendall() fast on PyPy3: 60MB/s -> 600MB/s

jamadden:
Share the chunking code between Python 2 and 3.

Also get the tests passing (read: skip a bunch) with 5.10.1 on OS X,
which is apparently less well tested than Linux.

We don't test pypy3 on appveyor, so:

[skip appveyor]

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -69,6 +69,9 @@
   it will regenerate itself. The default loop is the only one that can
   receive child events.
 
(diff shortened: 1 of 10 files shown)


jamadden:
Travis has been struggling mightily today, running extremely slowly and sometimes failing to show logs. There is a failure in the branch push, but I can't yet see why or tell if it's "routine" or not.

jamadden:
Builds are green.

PR #1109, opened 2018-02-19
Be more careful about issuing the SSL warning on Py2.

### LINKED ISSUE #1108: Unexpected SSL warning
* gevent version: 1.3a1
* Python version: 2.7.?
* Operating System: macOS

### Description:

I'm hearing reports that for some users, even though `from gevent import monkey; monkey.patch_all()` is the first line in their script, they still get the SSL warning introduced in #1016. That's unexpected and disturbing.

I'm seeking more details now. I know these users use buildout, so I'm wondering if the buildout wrapper script for their entry point is somehow leading to an SSL import?

jamadden:
Fixes #1108.

### DIFF
--- a/src/gevent/monkey.py
+++ b/src/gevent/monkey.py
@@ -103,14 +103,29 @@ class MonkeyPatchWarning(RuntimeWarning):
 saved = {}
 
 
-def is_module_patched(modname):
-    """Check if a module has been replaced with a cooperative version."""
-    return modname in saved
+def is_module_patched(mod_name):
+    """
+    Check if a module has been replaced with a cooperative version.
+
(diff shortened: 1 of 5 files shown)


src/gevent/monkey.py
```diff
@@ -295,6 +303,8 @@ def patch_thread(threading=True, _threading_local=True, Event=False, logging=Tru
                  existing_locks=True,
                  _warnings=None):
     """
+    patch_thread(threading=True, _threading_local=True, Event=False, logging=True, existing_lockes=True) -> None
```
arcivanov: existing_lock**E**s?

jamadden: good catch, thanks

PR #1134, opened 2018-03-07
Fix documentation for Pool.full()

parasyte:
The original docs had the purpose of the boolean return value reversed. I fixed that distinction, and also tried to clarify how the method can be used.


jamadden:
Thanks!

PR #1138, opened 2018-03-09
adding Supervisor functionality

Sevenops:
It would be nice if gevent implements Supervisor(https://hexdocs.pm/elixir/Supervisor.html ).
I know the test file is quite bad implemented i'm new at this area.

### DIFF
--- a/src/gevent/supervisor.py
+++ b/src/gevent/supervisor.py
@@ -1,0 +1,84 @@
+import time
+import gevent
+from collections import OrderedDict
+from enum import Enum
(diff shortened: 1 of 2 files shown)


jamadden:
Thanks for your PR! 

It's not clear to me that this belongs in gevent core, especially given how easy it is to implement without any knowledge of gevent internals. Perhaps you could release it as its own package to PyPI?

Sevenops:
ah okay thanks anyway.

PR #1158, opened 2018-03-30
WIP of Gevent-compatible MP

### LINKED ISSUE #993: Gevent and multiprocessing
* gevent version: 1.2.2
* Python version: Python 2.7.6
* Operating System: Ubuntu 14.04

### Description:

Tried running gevent based and multiprocessing based workers on the same codebase. The multiprocessing based worker blocks at start, probably related to this https://github.com/gevent/gevent/blob/master/src/gevent/monkey.py#L319-L340

### What I've run:

```
./manage.py celeryd -l debug -P gevent -c 100 --purge 
./manage.py celeryd -l debug -P processes -c 10 --purge 
```

Question: is this an issue being worked upon or should I just hack it through some global variables?

arcivanov:
fixes #993

### DIFF
--- a/src/gevent/_mp/2_7/__mp.py
+++ b/src/gevent/_mp/2_7/__mp.py
@@ -1,0 +1,70 @@
+import _multiprocessing as _mp
+import multiprocessing.reduction as mp_r
+import sys
+
+from gevent.hub import _get_hub_noargs as get_hub
+
+
+def recvfd(sockfd):
+    return get_hub().threadpool.apply(_mp.recvfd, (sockfd,))
+
+
+def sendfd(sockfd, fd):
+    return get_hub().threadpool.apply(_mp.sendfd, (sockfd, fd))
+
+
+SemLock = _mp.SemLock
+address_of_buffer = _mp.address_of_buffer
+flags = _mp.flags
+
+
+class Connection():
+    def __init__(self, handle, readable=True, writable=True):
+        self._conn = _mp.Connection(handle, readable, writable)
+
+    def poll(self, *args, **kwargs):
+        return get_hub().threadpool.apply(self._conn.poll, args, kwargs)
+
+    def recv(self):
+        return get_hub().threadpool.apply(self._conn.recv)
+
+    def recv_bytes(self, *args, **kwargs):
+        return get_hub().threadpool.apply(self._conn.recv_bytes, args, kwargs)
+
+    def recv_bytes_into(self, *args, **kwargs):
+        return get_hub().threadpool.apply(self._conn.recv_bytes_into, args, kwargs)
+
+    def send(self, *args, **kwargs):
+        return get_hub().threadpool.apply(self._conn.send, args, kwargs)
+
+    def send_bytes(self, *args, **kwargs):
+        return get_hub().threadpool.apply(self._conn.send_bytes, args, kwargs)
+
+    def close(self):
+        return get_hub().threadpool.apply(self._conn.close)
+
+    def fileno(self):
+        return self._conn.fileno()
+
+    def __repr__(self):
+        return self._conn.__repr__()
+
+    @property
+    def closed(self):
+        return self._conn.closed
+
+    @property
+    def readable(self):
+        return self._conn.readable
+
+    @property
+    def writable(self):
+        return self._conn.writable
+
+
+if sys.platform == 'win32':
+    PipeConnection = type("PipeConnection", (object,), dict(Connection.__dict__))
+    del Connection
+    mp_r.ForkingPickler.register(PipeConnection, mp_r.reduce_pipe_connection)
+else:
+    mp_r.ForkingPickler.register(Connection, mp_r.reduce_connection)
--- a/src/gevent/_mp/2_7/_mp_synchronize.py
+++ b/src/gevent/_mp/2_7/_mp_synchronize.py
@@ -1,0 +1,31 @@
+from multiprocessing.synchronize import SemLock as _SemLock, \
+    Semaphore as _Semaphore, \
+    BoundedSemaphore as _BoundedSemaphore, \
+    Lock as _Lock, \
+    RLock as _RLock
+
+from gevent.hub import _get_hub_noargs as get_hub
+
+__implements__ = ["SemLock", "Semaphore", "BoundedSemaphore", "Lock", "RLock"]
+__target__ = "multiprocessing.synchronize"
+
+
+class SemLock(_SemLock):
+    def _make_methods(self):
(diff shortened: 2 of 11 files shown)


arcivanov:
Work in progress - DO NOT MERGE

jamadden:
My initial thought is that this would be a good candidate for the "plugin" monkey-patching that's been brought up a few times earlier. Distributing this as a separate project (e.g., 'gevent_multiprocessing') at least initially, would free it from the gevent release cycle, giving it the easy ability to iterate quickly to fix early bugs, free it from having to support all the OS and Python combos gevent core does (e.g., Windows is an issue in this right now), and so on.

I was hoping to get gevent 1.3b1 out this weekend, since Python 3.7b3 dropped, but that's been pushed back to Monday or Tuesday by circumstances. Adding a simple plugin system won't take me too long. I'm roughly thinking setuptools entry points, adjust `patch_all` to take **kwargs, which get passed to the discovered entry points, document the patching functions.

What do you think about that?

arcivanov:
Are you thinking extra-based plugin? Or a separate project altogether?

arcivanov:
I think plugin system should be fine, although a counter-argument to that is that if Gevent monkey-patch tries to make Python-proper cooperative, both MP and asyncio are part of the core specifications and therefore putting them in separate plugins would be counter-intuitive.

arcivanov:
So another very important issue I've discovered just now that this construct becomes necessary in cases where we don't want the child process (e.g. ForkServer) to accidentally load gevent (MP starts the child process with a main module of a parent, which results in gevent monkey-patch becoming dragged into unwantingly):

```
from multiprocessing.process import current_process

from gevent import monkey

if not getattr(current_process(), "_inheriting", False):
    monkey.patch_all()
```

If multiprocessing is a module then such conditions should be pluggable at the top of patch_all to make sure patching can be disabled.

jamadden:
> a separate project altogether?

That's definitely the way I'm leaning, at least initially, for the pragmatic reasons I mentioned. The standard library often takes the approach of saying "but it on PyPI, see what the popularity and maintenance concerns are like before making it the responsibility of the core" and I can see that being a reasonable tactic here.

> both MP and asyncio are part of the core specifications and therefore putting them in separate plugins would be counter-intuitive.

That's certainly a valid point. The counter-counter argument is that gevent is "a coroutine-based Python networking library" (documentation) whose primary objective is to enable network based servers (me). "greening-all-the-things" is a (useful) secondary goal, and often a necessary one. In my experience, multiprocessing is not very commonly used by networking programs and servers; servers tend to use an external task queue of some sort. It looks like it takes a large surface area and moderately invasive patching (e.g., `__module__` swizzling) to make multiprocessing green. So also from those two points a separate library (again, at least initially) seems reasonable to me.

We would of course want to link to this library from the documentation, and since the failure case of patch_threading + multiprocessing is so bad (hung process) we might want to look into a way to issue a runtime warning as part of core if we can detect that scenario.

(asyncio is a separate ball of wax. 😄)

> such conditions should be pluggable at the top of patch_all to make sure patching can be disabled.

Reasonable. My current thoughts are to combine setuptools entry points and the new gevent.events module to "broadcast" important steps in the patching process to both plugins and event subscribers. If any receiver raised a special exception, the process would stop.

Roughly:

```python
def patch_module(module):
   try:
      broadcast_gevent_will_patch_module(module)
   except DoNotPatch:
      return False
   ...
   broadcast_gevent_did_patch_module(module)

def patch_all(**kwargs):
   try:
     broadcast_gevent_will_patch_all(kwargs)
   except DoNotPatch:
     return False
   
   ... 

   execute_patch_plugins(kwargs)
   broadcast_gevent_did_patch_all(kwargs)
```

arcivanov:
Let me process this a bit. 

Meanwhile, there is a whole bunch of failures in 3.5 and 3.4:

1/125 unexpected failures
 - GEVENT_DEBUG=debug GEVENT_RESOLVER=thread GEVENT_RESOLVER_NAMESERVERS=8.8.8.8 /home/travis/.runtimes/snakepit/python3.5.4 -u test__local.py

 - GEVENT_DEBUG=error GEVENT_RESOLVER=thread GEVENT_RESOLVER_NAMESERVERS=8.8.8.8 /home/travis/.runtimes/snakepit/python3.4.7 -u -W ignore -mgreentest.monkey_test test_threading.py
 - GEVENT_DEBUG=debug GEVENT_RESOLVER=thread GEVENT_RESOLVER_NAMESERVERS=8.8.8.8 /home/travis/.runtimes/snakepit/python3.4.7 -u test__local.py

jamadden:
While you process, I’d like to clarify that I absolutely *do* feel this is a significant piece of work and I *do* believe it of value to the gevent community. I am also hopeful it can serve as a pilot project in a  way to share parts of development and maintenance throughout the community.

jamadden:
> Meanwhile, there is a whole bunch of unrelated failures in 3.5 and 3.4:

Hmm. There are a few timing related failures I have learned can often be solved by simply rerunning the affected scenario in isolation, but spot checking a few of those didn’t look the same. So I don’t know what’s up. I would suggest (and I know this is weak tea) running each scenario individually again.

arcivanov:
> There are a few timing related failures

It seems that the failures are ample even in master.

arcivanov:
> While you process, 
I didn't take it as if you didn't. :smile:  Thanks!

jamadden:
> It seems that the failures are ample even in master.

Hmm, I don't see what you're talking about. The most recent master builds on travis [are green](https://travis-ci.org/gevent/gevent/builds/361189474) as was Appveyor for Python 3. I don't reproduce any failures with 3.5 locally on master either.

jaddison:
This would be amazing to see completed - what is missing/outstanding for this to be made available?

jamadden:
gevent 1.3 should contain all the necessary hooks to implement this as the suggested separate library.

arcivanov:
Sorry guys, I'll get back to it soon.

maingoh:
Just curious, @arcivanov are you planning to finish it ? 
I also have a question, what are the benefits of this PR compared to https://github.com/jgehrcke/gipc ?
Thank you !

arcivanov:
@maingoh 
1) Yes, although if you read through the comments, it will be a separate project pluggable into Gevent via extensions API.
2) This, in theory, would make **all** of the features of regular Python `mp` "just work" OOB. Just like geven monkey-patching would otherwise :)

maingoh:
Thank you for the answers ! Looking forward to see this merged :smiley:

arcivanov:
This PR is going away, please see https://github.com/karellen/geventmp

PR #1169, opened 2018-04-03
Add extension points to gevent.monkey using events and setuptools entry points.

### LINKED ISSUE #1162: Monkeypatch io.read/io.write/open/io.open to track blocking calls in debug mode
This is related to #1070 and provides an safe debug-time alternative.

Without making any of the above methods non-blocking/loop-friendly, it might be useful to provide facilities to mp all blocking IO methods in debug/troubleshooting mode to track blocking IO operations. 

Even if the mp'ed code functions correctly in test, it may be only due to circumstances surrounding the environment, i.e. if a method accepts a domain socket and there are no buffer overflows the loop would be responsive simply due to localhost communication even if blocking IO is on a critical path. Same code where unix socket is replaced with a TCP one talking remotely over a slow link will grind the loop to a halt, which in turn is very hard to troubleshoot. 

By introducing an ability to optionally mp all blocking IO code without modifying its behavior (other than auditing) would provide insight into missed mp branches.

jamadden:
Also use the same mechanism for adding new functions to the monitoring thread.

Fixes #1162. Refs #1158.


jamadden:
@arcivanov does this give the necessary hooks? If so I would like to merge and release.

jamadden:
Aww snap, I think I forgot to add **kwargs or equivalent to patch_all

jamadden:
I have a family emergency. I may be out of touch for some days

arcivanov:
Best wishes!

PR #1176, opened 2018-04-15
Use /dev/fd|/proc/self/fd to get open FDs to close in Popen

### LINKED ISSUE #1172: _close_fds on linux takes relatively long time on Python 3
https://github.com/gevent/gevent/blob/e3e555e2309120a488bbbc8b4f965791a5f15eed/src/gevent/subprocess.py#L1148

This is true especially when the host has a high limit of file-descriptors (`MAXFD`), which seems to be the case inside docker containers. This implementation leads to a lot (~2^20 in my case) of unnecessary calls to the `close` syscall, which slows down any subprocess invocation.
In contrast, the native python implementation uses `_posixsubprocess.fork_exec` [which in turn uses](https://github.com/python/cpython/blob/master/Modules/_posixsubprocess.c#L273) `/dev/fd` to identify fds that are indeed open.

jamadden:
If those aren't available, use the old brute-force approach. This is closer to what CPython does in its C implementation, and is much faster. The speed was bade enough I consider the cases to be a regression on Python 3.

We don't have to worry about the async signal safe stuff the C code does because, guess what, we're running Python code here already anyway, so much of it could wind up doing something that's not actually safe anyway. Oh well.

Since we depend on Python 3.4 and above now, we can rely on the CLOEXEC flag being set by default and not have to manually check everything.

This speeds up 2.7 (close_fds defaults to *false* there, so the default case doesn't change):

| Benchmark                 | 27_bench_subprocess | 27_bench_subprocess_dirfd     |
|---------------------------|---------------------|-------------------------------|
| spawn native no close_fds | 1.81 ms             | 1.79 ms: 1.01x faster (-1%)   |
| spawn gevent no close_fds | 2.11 ms             | 2.20 ms: 1.04x slower (+4%)   |
| spawn native close_fds    | 31.0 ms             | 30.2 ms: 1.03x faster (-3%)   |
| spawn gevent close_fds    | 31.6 ms             | 2.56 ms: 12.31x faster (-92%) |

And it really speeds up 3.7 (close_fds defaults to *true* there, so the default case is much faster, and the non-default case is even better):

| Benchmark                 | 37_bench_subprocess | 37_bench_subprocess_dirfd     |
|---------------------------|---------------------|-------------------------------|
| spawn native no close_fds | 1.34 ms             | 1.27 ms: 1.06x faster (-6%)   |
| spawn gevent no close_fds | 117 ms              | 3.05 ms: 38.27x faster (-97%) |
| spawn native close_fds    | 1.36 ms             | 1.30 ms: 1.04x faster (-4%)   |
| spawn gevent close_fds    | 32.5 ms             | 3.34 ms: 9.75x faster (-90%)  |

Fixes #1172

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,6 +10,18 @@
 - On Windows, CFFI is now a dependency so that the libuv backend
   really can be used by default.
 
+- Fix a bug detecting whether we can use the memory monitoring
+  features when psutil is not installed.
+
+- `gevent.subprocess.Popen` uses ``/proc/self/fd`` (on Linux) or
+  ``/dev/fd`` (on BSD, including macOS) to find the file descriptors
+  to close when ``close_fds`` is true. This matches an optimization
+  added to Python 3 (and backports it to Python 2.7), making process
+  spawning up to 9 times faster. Also, on Python 3, since Python 3.3
+  is no longer supported, we can also optimize the case where
+  ``close_fds`` is false (not the default), making process spawning up
+  to 38 times faster. Initially reported in :issue:`1172` by Ofer Koren.
+
 
 1.3b1 (2018-04-13)
 ==================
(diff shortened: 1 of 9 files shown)


jamadden:
Hmm, I think I'd like to do some work to make those functions more testable independently.

PR #1191, opened 2018-04-23
RLock: Added method locked

hackaugusto:


jamadden:
Thanks! Can you describe a motivating use case? That class is intended to provide the same API as the stdlib version, and I don’t see that method in [the docs](https://docs.python.org/3.7/library/threading.html#rlock-objects).

arcivanov:
I would further argue that the method is useless and extremely dangerous. Locked status can only be used for debug purposes and even as such it is unreliable. 
No code can be correct if used:
```python
if lock.locked():
    do_something()
```
as the above is a critical section that must hold a lock to be correct in the first place.

hackaugusto:
> Thanks! Can you describe a motivating use case?

```python
if lock.locked():
  raise WouldBlock()
with lock:
  # do some operation that must be executed only once
  ...
```

I had this code with a binary semaphore. It was written as such mainly to use the context manager for cleaning up. Once the lock changed from `Semaphore` to `RLock` it stopped working because of the interface differences.

The above snippet can be rewritten as:

```python
if lock.acquire(blocking=False):
  # do the operation
  ...
  lock.release()
else:
  raise WouldBlock()
```

The motivation of the PR is mainly to have the same interface as `Semaphore` and allow the use of the context manager for cleanup.

arcivanov:
```
if lock.locked():
  raise WouldBlock()
with lock:
  # do some operation that must be executed only once
  ...
```
This is broken logic. Please don't use it. This idiom is broken in EVERY language and VM (C/++, JVM) where it is attempted. And the rewrite with `try_acquire`/`acquire(blocking=False)` is fundamentally different - those two code blocks **are not** equivalent. 

https://softwareengineering.stackexchange.com/questions/330769/why-cant-i-check-whether-a-mutex-is-locked?utm_medium=organic&utm_source=google_rich_qa&utm_campaign=google_rich_qa

This is why the idiom you're requesting does NOT appear in the RLock in the first place or any other language.

jamadden:
@hackaugusto I could imagine usage of this API for debugging purposes, and symmetry with Semaphore could be a compelling argument, but @arcivanov is correct, using this API for "double checked locking" is considered a broken pattern. That's not something that we want to encourage.

One could argue that we know that we're operating in a single thread and we know exactly when preemption could occur (i.e., it can't, if we're careful) so the brokenness isn't *really* broken. Because Semaphore is a gevent-specific API, I can accept that argument there, but RLock follows the standard library and it's much less likely that users would consider all the factors that go into correct usage there.  There are better, safer, correct alternatives. The points about multiple greenlets only make the intended use of this API even muddier.

As such, I don't think we can accept this PR.

hackaugusto:
> As such, I don't think we can accept this PR.

Fair enough. Perhaps you should consider deprecating the API from the Semaphore class.

> This is broken logic. Please don't use it. This idiom is broken in EVERY language and VM (C/++, JVM) where it is attempted

Thanks for the warning, but could you explain why this logic is broken on gevent? (I assume CPython + GIL + cooperative scheduling on IO)

Issue #1126, opened 2018-03-01
libuv check watcher infinite loop


jamadden:
Running CPython 3.7b2 on macOS, I can sometimes produce an infinite loop where a single check watcher gets called over and over again. `python -m greentest.monkey_test test_selectors.py PollSelectorTestCase.test_timeout` will reliable produce this (with 'test_main()' replaced with 'unittest.main()' so we can call a single test).

That test creates a number of poll watchers and a single check watcher (a Timeout with time of 0, which we turn into a check watcher to make the loop work). 

Here's some debugging output from a modified libuv with comments:
```console
Entering run for check # begin uv__run_check
Calling watcher check 0x7ff67d4bbe30 # Our main callback watcher
Stopping for check 0x7ff67d4bbe30; is active? 1
Stopping for check 0x7ff67d4be9d0; is active? 1
UV_CHECK_CLOSE 0x7ff67d4be9d0 # Yep, stopped *and* closed
Stopping for check 0x7ff67d4be9d0; is active? 0
Starting for check 0x7ff67d4e91b0; is active? 0 # WATCH THIS ONE: this is the timer 0
Calling watcher check 0x7ff67d4be9d0 # WAIT, WHOA: You were already stopped!
Calling watcher check 0x7ff67d4bb780
Exiting run for check # Exit that check
Entering run for idle
Exiting run for idle
Entering run for prepare
Calling watcher prepare 0x7ff67d4bb900
Stopping for check 0x7ff67d4bbe30; is active? 0
Exiting run for prepare
Starting for check 0x7ff67d4bbe30; is active? 0

Entering run for check  # Begin another uv__run_check
Calling watcher check 0x7ff67d4bbe30 # main callback again
Stopping for check 0x7ff67d4bbe30; is active? 1
Stopping for check 0x7ff67d4e91b0; is active? 1 # We stop it
UV_CHECK_CLOSE 0x7ff67d4e91b0 # And we close it
Stopping for check 0x7ff67d4e91b0; is active? 0
Starting for check 0x7ff67d4be9d0; is active? 0
Calling watcher check 0x7ff67d4e91b0
WARNING: gevent: Unable to dereference handle; not stopping watcher. Native resources may leak. This is most likely a bug in gevent.
Calling watcher check 0x7ff67d4be9d0
Stopping for check 0x7ff67d4be9d0; is active? 1
UV_CHECK_CLOSE 0x7ff67d4be9d0
[...]

PR #1194, opened 2018-04-25
[WIP] Queue all libuv callbacks and run them at the end of the iteration

jamadden:
This is an alternate fix to #1126 as suggested on the mailing list.

It seems to work, but it does have the problem that prepare and idle watchers are not run until after we poll for IO, not before. Likewise, it delays timers (on unix, libuv runs timers both before and after polling for IO). This didn't cause any test failures, but I don't like it. It's easy to see that it introduces artificial delays to timers.

I'm not sure what to do about it. Ideally we could interject code to run just before `uv__io_poll` and *really* run the timers/idle/prepare watchers. That's what prepare is for, but we can't actually run any code in the prepare watcher. One idea is to implement prepare/idle/timer watchers in python and manually call them as we iterate the loop. That sounds expensive...but on the plus side it might let us get better timer resolution?


jamadden:
A partial solution might be to run the queued callbacks when we add a new callback that's a different type than the last queued callback:

```python
def _queue_callback(self, watcher_ptr, revents):
   if self._queued_callbacks and self._queued_callbacks[-1].type != watcher_ptr.type:
      self.__run_queued_callbacks()
   self._queued_callbalks.append((watcher_ptr, revents))
```

In effect this lets us run code "in between the lines" inside `uv_run` as it moves from one type of watcher to the next. We'd be assured that each individual function (e.g., `uv__run_prepare`) has finished by the time we run prepare callbacks.

Part of the loop:
```c
  while (r != 0 && loop->stop_flag == 0) {
    uv__update_time(loop);
    uv__run_timers(loop);
    ran_pending = uv__run_pending(loop);
    uv__run_idle(loop);
    uv__run_prepare(loop);
```

Of course, if an idle watcher wanted to switch greenlets while also modifying the prepare queue, we'd be right back in the same boat, just harder to debug...

jamadden:
This snippet demonstrates the problem:
```python
l = get_hub().loop

from perf import perf_counter

io = l.io(1, 1)
def icb(evencts):
	print("icb")
io.start(icb)

now = perf_counter()
t = l.timer(0.0001)
def cb():
	then = perf_counter()
	print("Timer fired after %1.6f" % (then  - now))

t.start(cb)

get_hub().join()
```

With some debugging added, we get this output:
```
Queue cb <cdata 'struct uv_handle_s *' 0x7fdee9665410> 0.108916224
Using timeout 265 (ran pending? 0)
Running cb for <cdata 'struct uv_handle_s *' 0x7fdee9665410> delay 0.269353
Timer fired after 0.270677
```

libuv is letting the call to select() sleep for 265 ms, which is how long it takes for our timer to actually get called. 

If there wasn't an active IO watcher, the timer would be called at the appropriate time because libuv would choose a select() timeout of 0---I suspect that might explain why all the tests pass; there are a large variety of circumstances in which libuv will choose a poll timeout of 0:
```c
int uv_backend_timeout(const uv_loop_t* loop) {
  if (loop->stop_flag != 0)
    return 0;

  if (!uv__has_active_handles(loop) && !uv__has_active_reqs(loop))
    return 0; // we get here if the io watcher is never started.

  if (!QUEUE_EMPTY(&loop->idle_handles))
    return 0;

  if (!QUEUE_EMPTY(&loop->pending_queue))
    return 0;

  if (loop->closing_handles)
    return 0;

  return uv__next_timeout(loop); // the time for the next timer
}
```

jamadden:
You might think it odd that the timeout is calculated as 265 instead of 1, since 1 is the next minimum timer in the timer heap. That certainly confused me when I looked at it a bit more closely.

What's happening is that libuv runs timers *twice* (in UV_RUN_ONCE mode), once before polling for IO, and once after. When a timer runs, if it's not set to repeat, it is stopped, taking it out of the timer heap, so it is not used in the calculation of `uv__next_timeout`.

I've extracted the snippet to a test case, one of three or four tests in a file. If that test runs normally along with the others, I don't reproduce the issue, but if I run it independently I do. This is despite the fact that each test uses a fresh `uv_loop` object. Still not sure what's going on there...

Issue #1016, opened 2017-09-11
SSLContext infinite recursion in Python 3.6


Teyras:
On Python 3.6.2 and gevent 1.2.2, I encounter the problem described in #903 . I do not have the rights to reopen the issue, which is why I'm making a duplicate.

jamadden:
Can you please post a failing example and its output?

Teyras:
The code:

```
import gevent
from gevent import monkey
import requests

monkey.patch_all()
from requests.packages.urllib3.util.ssl_ import create_urllib3_context
create_urllib3_context()
```
The output:
```
Traceback (most recent call last):                                                        
  File "gevent_requests.py", line 7, in <module>                                          
    create_urllib3_context()
  File "/usr/lib/python3.6/site-packages/urllib3/util/ssl_.py", line 269, in create_urllib
3_context             
    context.options |= options
  File "/usr/lib/python3.6/ssl.py", line 459, in options
    super(SSLContext, SSLContext).options.__set__(self, value)                            
  File "/usr/lib/python3.6/ssl.py", line 459, in options
    super(SSLContext, SSLContext).options.__set__(self, value)
  File "/usr/lib/python3.6/ssl.py", line 459, in options
    super(SSLContext, SSLContext).options.__set__(self, value)
  [Previous line repeated 329 more times]    
RecursionError: maximum recursion depth exceeded while calling a Python object
```

jamadden:
Thanks. The problem is that you're importing `requests` *before* you monkey patch. [You must monkey patch](http://www.gevent.org/gevent.monkey.html#patching) before [importing anything else](http://www.gevent.org/intro.html#monkey-patching). I can reproduce this if I patch in the incorrect order, but if I patch in the correct order it works as expected:
```pycon
Python 3.6.2 (default, Jul 18 2017, 14:08:57)
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.42)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import gevent.monkey
>>> gevent.monkey.patch_all()
>>> from requests.packages.urllib3.util.ssl_ import create_urllib3_context
>>> create_urllib3_context()
<gevent._ssl3.SSLContext object at 0x1013c2228>
```

Teyras:
Yes, that seems to be the problem - after fixing the order, everything is fine. I'm sorry for bothering you with this, I just googled the exception, found the issue I referenced and thought this must be the problem.

[...]

PR #1239, opened 2018-06-15
Only issue the ssl MonkeyPatchWarning if we can be pretty sure there will be problems

jamadden:
This increases the false negative chances, but decreases the false positives substantially.

Given 
```python
import requests
import gevent.monkey; gevent.monkey.patch_all()
```

The expected warning would be something like:

> MonkeyPatchWarning: Monkey-patching ssl after ssl has already been imported may lead to errors, including RecursionError on Python 3.6. It may also silently lead to incorrect behaviour on Python 3.7. Please monkey-patch earlier. See https://github.com/gevent/gevent/issues/1016. Modules that had direct imports (NOT patched): ['urllib3.util (//lib/python3.6/site-packages/urllib3/util/__init__.py)', 'urllib3.util.ssl_ (//lib/python3.6/site-packages/urllib3/util/ssl_.py)'].

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,7 +7,10 @@
 1.3.4 (unreleased)
 ==================
 
-- Nothing changed yet.
+- Be more careful about issuing ``MonkeyPatchWarning`` for ssl
+  imports. Now, we only issue it if we detect the one specific
+  condition that is known to lead to RecursionError. This may produce
+  false negatives, but should reduce or eliminate false positives.
 
 
 1.3.3 (2018-06-08)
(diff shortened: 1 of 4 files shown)


jamadden:
The "incorrect behaviour on Python 3.7" is due to the addition of `SSLContext.sslsocket_class`, which is used by `SSLContext.wrap_socket`.  [It is documented](https://docs.python.org/3.7/library/ssl.html#ssl.SSLContext.sslsocket_class) to accept a subclass of `SSLSocket`. It is used by `wrap_socket` to create sockets, but `wrap_socket` calls a private, undocumented `_create` method on it which our sockets do not have. This was changed in 3.7 when it was declared that the constructor of `SSLSocket` was private.

For this reason, we could not monkey-patch `SSLContext` in-place by changing out its `sslsocket_class` to our own. If nothing ever touches the broken properties (unclear if anything does), then you could silently be using the wrong types of sockets.

Arguably, gevent should adopt `_create` and a private constructor for the `SSLSocket` on 3.7, but not much seems to depend on that so I hadn't gone to the trouble yet.

PR #1240, opened 2018-06-16
Use a buffer to generate HTTP chunks.

### LINKED ISSUE #1233: pywsgi.py with SSE produces extraneous TCP packets
* gevent version: 1.2.2
* Python version: Python 2.7.14
* Operating System: Ubuntu x86_64

### Description:
Using various versions of SSE (Server Side Events), for example from here: http://flask.pocoo.org/snippets/116/ with the pywsgi embedded server, running Wireshark shows multiple packets per server push. This is caused by:

https://github.com/gevent/gevent/blob/master/src/gevent/pywsgi.py#L724-L726

which itself is caused by never running https://github.com/gevent/gevent/blob/master/src/gevent/pywsgi.py#L698 . 

A browser will ignore the extra packets, but they are sent and processed anyway. I could not find an easy workaround by changing settings or headers. Using gunicorn to serve, rather than the build-in WSGIServer does not produce these extra packets.

jamadden:
The intent is to reduce network overhead for small chunk sizes withoud adding much overhead for larger sizes.

Fixes #1233.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,9 @@
   condition that is known to lead to RecursionError. This may produce
   false negatives, but should reduce or eliminate false positives.
 
+- Based on measurements and discussion in :issue:`1233`, adjust the
+  way :mod:`gevent.pywsgi` generates HTTP chunks. This is intended to
+  reduce network overhead, especially for smaller chunk sizes.
 
 1.3.3 (2018-06-08)
 ==================
--- a/src/gevent/pywsgi.py
+++ b/src/gevent/pywsgi.py
@@ -708,7 +708,9 @@ class WSGIHandler(object):
             raise
         self.response_length += len(data)
 
-    def _write(self, data):
+    def _write(self, data,
+               _PY34_EXACTLY=(sys.version_info[:2] == (3, 4)),
+               _bytearray=bytearray):
         if not data:
             # The application/middleware are allowed to yield
             # empty bytestrings.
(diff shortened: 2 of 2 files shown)


arcivanov:
You're exchanging memory allocation/collection pressure for non-blocking syscall latency. Does it make sense for performance/scalability?

jamadden:
It's a little unpredictable because of the way `sendall` may expand into many `send` calls and many more data slices. On CPython, the slicing that's done for sending bytearrays is effectively free, we only pay the cost of the concatenation. Based on the data in #1233, it seems that the concatenation costs less than the extra syscalls/packets. 

At concurrency 10, which was as high as I tested, my gevent server was fully utilizing the CPU for *all three cases*: non-chunked (single sendall call, arbitrary `send` syscalls), chunked with separate writes (the status quo: three sendall calls, arbitrary `send` syscalls) and this PR, chunked with concatenation (single sendall call, arbitrary `send` syscalls). Overall throughput was not as good for chunked as non-chunked with this PR, but still *dramatically* better for chunked than it was before, at the same CPU utilization.

The cost is a (bounded and time-limited) extra memory usage. Since I expect most chunks to be small (based on just guessing!), it seems worth it to me. (Hmm, a slight modification of this PR might allow reference counting to reclaim some memory a bit sooner than otherwise by setting `data = None` before the sendall call. How I wish Python had tail-call optimizations 😄 Also, this is a place that libuv might theoretically help us out; it has a portable 'sendmsg' equivalent in its `uv_tcp_t` type that accepts multiple buffers. )

I am open to other ideas, of course.

PR #1241, opened 2018-06-18
Use bytearray += instead of bytearray.extend

jamadden:
This compiles to fewer bytecodes and uses built-in slots instead of method dispatch, so it should be faster.

Benchmarks show a tiny improvement (using the same benchmarks as issue #1233, so not exactly designed to test this):

RPS

|Size  |master|+=     |
|------|------|-------|
|1     |5.60  | 5.63  |
| 10   |5.49  |  5.53 |
|  100 |4.99  |   5.00|
|  1000|3.53  |   3.57|

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,9 @@
   way :mod:`gevent.pywsgi` generates HTTP chunks. This is intended to
   reduce network overhead, especially for smaller chunk sizes.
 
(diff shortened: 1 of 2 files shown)


jamadden:
The Travis PR failure is a git error of some sort, probably having to do with a corrupted cache of some sort.

PR #1255, opened 2018-07-11
Enable 3.7 on appveyor.

jamadden:


jamadden:
There are test failures on Appveyor, but the last two builds with 3.7 have been clean; the test failures are the "normal" racy, load-related failures.

There are test failures on Travis, but this changed nothing for Travis. The failures there are also the normal, racy, load-related failures, coupled with the remaining fallout from the purging and rebuilding of the caches.

PR #1278, opened 2018-09-12
fixed negative length in pywsgi's Input read functions for non chunked body

tzickel:


jamadden:
Thanks! Travis is green. The Appveyor failure is an intermittant one I'm not worried about. The only thing left is testing the readline version, as mentioned over in the original issue, and updating the changelog. Do you feel like adding that?

tzickel:
Done.

PR #1284, opened 2018-10-11
Update to libuv 1.23.2

jamadden:


jamadden:
Travis is green. Neither of the appveyors were green, but they each failed on a completely different version of Python, with one of the usual timeout-ish errors. I think we can merge this.

PR #1288, opened 2018-10-15
Begin WaitIterator iteration when __next__ is called.

hashbrowncipher:
For scenarios where they only want to get the first greenlet which exits, a
user might write this code:

    next(gevent.iwait(greenlets))

Without this commit, the above snippet will hang, because the waited-upon
objects have not been `rawlink()`ed. This change makes such bugs impossible.


hashbrowncipher:
I've made the requested changes and rebased my branch to eliminate conflicts.

jamadden:
Great, thanks!

PR #1289, opened 2018-10-15
Add Greenlet.add_spawn_callback() and Greenlet.remove_spawn_callback()

1st1:
These methods allow to set up code to intercept Greenlet objects
creation.  This is useful to implement custom tracing of gevent tasks
and custom variants of `gevent.local()`.

The performance impact is minuscule: one extra fast "if" check in
`Greenlet.__init__`.

### DIFF
--- a/src/gevent/greenlet.py
+++ b/src/gevent/greenlet.py
@@ -508,6 +508,7 @@ class Greenlet(greenlet):
     def start(self):
         """Schedule the greenlet to run in this loop iteration"""
         if self._start_event is None:
+            _call_spawn_callbacks(self)
             self._start_event = self.parent.loop.run_callback(self.switch)
 
     def start_later(self, seconds):
@@ -518,6 +519,7 @@ class Greenlet(greenlet):
         *seconds* later
         """
         if self._start_event is None:
+            _call_spawn_callbacks(self)
             self._start_event = self.parent.loop.timer(seconds)
             self._start_event.start(self.switch)
 
@@ -572,7 +574,6 @@ class Greenlet(greenlet):
             instead of spawning a greenlet that will raise an uncaught TypeError.
         """
         g = cls(*args, **kwargs)
-        _call_spawn_callbacks(g)
         g.start()
         return g
 
@@ -596,7 +597,6 @@ class Greenlet(greenlet):
         if cls is Greenlet and not args and 'run' not in kwargs:
             raise TypeError("")
         g = cls(*args, **kwargs)
-        _call_spawn_callbacks(g)
         g.start_later(seconds)
         return g
 
--- a/src/gevent/greenlet.py
+++ b/src/gevent/greenlet.py
@@ -538,7 +538,7 @@ class Greenlet(greenlet):
         .. versionadded:: 1.3.8
         """
         global _spawn_callbacks
-        if _spawn_callbacks is None:
+        if _spawn_callbacks is None:  # pylint:disable=used-before-assignment
             _spawn_callbacks = set()
         _spawn_callbacks.add(callback)
 
@@ -929,10 +929,10 @@ def _killall(greenlets, exception):
                 g.parent.handle_error(g, *sys_exc_info())
 
 
-def _call_spawn_callbacks(greenlet):
+def _call_spawn_callbacks(gr):
     if _spawn_callbacks is not None:
         for cb in _spawn_callbacks:
-            cb(greenlet)
+            cb(gr)
 
 
 _spawn_callbacks = None
--- a/src/gevent/greenlet.py
+++ b/src/gevent/greenlet.py
@@ -264,9 +264,6 @@ class Greenlet(greenlet):
             self.spawn_tree_locals = None
             self._spawning_stack_frames = None
 
-        if _spawn_callbacks is not None:
-            self.__call_spawn_callbacks()
-
     @Lazy
     def spawning_stack(self):
         # Store this in the __dict__. We don't use it from the C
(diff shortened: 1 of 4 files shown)


1st1:
Context: we use gevent heavily at Pinterest. Having this functionality in gevent would preclude us from maintaining a private fork.  We also believe that this functionality is generally useful for any large-scale gevent deployment with custom instrumentation to trace tasks/io and persist task-local data.

jamadden:
Hi, and thanks for the contribution!

One question I have is why couldn't this be done with a subclass in your code base? What I mean is, is there someplace in gevent or elsewhere that specifically and only uses a hardcoded `Greenlet.spawn` or `Greenlet()` call? gevent generally tries to make using subclasses easy with things like the [`greenlet_class` attribute of a Pool](http://www.gevent.org/api/gevent.pool.html#gevent.pool.Group.greenlet_class) and similar; if there are places that are missing we can work on adding them.

I'm somewhat concerned by the way the reference to `self` now escapes `__init__` potentially before subclass `__init__` have run and finished their set up. What if the callbacks were called from `start()` and `start_later()` instead?

1st1:
> One question I have is why couldn't this be done with a subclass in your code base? What I mean is, is there someplace in gevent or elsewhere that specifically and only uses a hardcoded Greenlet.spawn or Greenlet() call?

Using `Greenlet()` directly to create tasks is a [documented pattern](http://www.gevent.org/api/gevent.greenlet.html#starting-greenlets) and we have some examples of it in our (constantly growing) code base.  I suspect that it's probably used in some third-party libraries as well.  We simply can't be sure that some library that has always been using `spawn()` wouldn't start using `Greenlet()` at some point.

Therefore the most reliable option for us was monkey patching of the `gevent.Greenlet.__init__` function.  That way we were absolutely certain that all `gevent.Greenlets` (no matter how they were created) are instrumented and we always get complete traces.  However monkey patching, as of gevent 1.3.x, no longer works, because `Greenlet` type is now compiled with Cython and `Greenlet.__init__` is now readonly.

FWIW, this use case is exactly why [`asyncio.set_task_factory()`](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.set_task_factory) exists.

> I'm somewhat concerned by the way the reference to self now escapes __init__ potentially before subclass __init__ have run and finished their set up. What if the callbacks were called from start() and start_later() instead?

Doing that in `start()` and `start_later()` would also work!

src/gevent/greenlet.py
```diff
@@ -521,6 +529,41 @@ def start_later(self, seconds):
...
 
+    @staticmethod
+    def add_spawn_callback(cb):
+        """
+        add_spawn_callback(callback) -> None
+
+        Set up a *callback* to be invoked when a new :class:`Greenlet`
+        object is instantiated.
+
+        The invocation order of spawn callbacks is unspecified.  Adding one
```
jparise: "Adding a callback more than once ..." might be even clearer here. What do you think?

src/gevent/greenlet.py
```diff
@@ -521,6 +529,41 @@ def start_later(self, seconds):
...
+            _spawn_callbacks = set()
+        _spawn_callbacks.add(cb)
+
+    @staticmethod
+    def remove_spawn_callback(cb):
+        """
+        remove_spawn_callback(callback) -> None
+
+        Remove *callback* function added with :meth:`Greenlet.add_spawn_callback`.
+        This function will not fail if *callback* has been already removed.
```
jparise: Or if the callback was never added.

1st1:
I've updated the PR to run callbacks in `Greenlet.start()` and `Greenlet.start_later()`.  The CI should also be fixed now. 🤞

jamadden:
> I've updated the PR to run callbacks in Greenlet.start() and Greenlet.start_later(). The CI should also be fixed now. 

It looks like it actually is doing it from `Greenlet.spawn()` and `Greenlet.spawn_later()`. Is that what you meant? That means that `Greenlet(a_function).start()` will not call the callbacks.

jamadden:
> That way we were absolutely certain that all gevent.Greenlets (no matter how they were created) are instrumented and we always get complete traces. 

Thanks, that makes sense and is pretty much what I was thinking.

1st1:
> It looks like it actually is doing it from Greenlet.spawn() and Greenlet.spawn_later(). Is that what you meant? That means that Greenlet(a_function).start() will not call the callbacks.

Good catch!  Yes, I wanted to change `start()` and `start_later()` and was confused by similarly looking `spawn()` and `spawn_later()`.  I've pushed another commit with improved unittests to make sure we test `Greenlet().start()`, `Grenlet.spawn()`, and `Greenlet.spawn_later()`.

src/greentest/test__greenlet.py
```diff
@@ -699,6 +699,39 @@ def func():
         while not raw.dead:
             gevent.sleep(0.01)
 
+    def test_add_spawn_callback(self):
+        called = 0
+
+        def cb(gr):
+            nonlocal called
```
jamadden: We support Python 2 still, where [nonlocal is a syntax error](https://travis-ci.org/gevent/gevent/jobs/441913993#L678).

1st1: Right. I've pushed a fix.

1st1:
Travis is green; AppVeyor has one failed test that seems to be unrelated to this PR.

jamadden:
I agree the failure is unrelated. Thanks for the PR!

1st1:
Hi @jamadden, is there any ETA for the next release that includes this change?

jamadden:
No specific ETA, but hopefully before end-of-year. (I'm pretty slammed right now with other commitments.) I'm also hoping manylinux2010 comes out so I can distribute libuv support in the binary linux wheels.

1st1:
Hi @jamadden!  Would it be possible to issue a release with this change so that we can start using the upstream gevent?

jamadden:
Done. I missed EOY deadline, but manylinux2010 doesn't look like it's going to be officially ready for primetime very soon.

1st1:
Thanks a lot, Jason!

PR #1290, opened 2018-10-16
WaitIterators: only return a given object once

hashbrowncipher:
Depending on feedback, this PR may supersede #1288.

There are detailed commit messages describing what is going on. As a summary, this PR:

* documents a pitfall of `gevent.wait`

Other than documentation, the alternative would be to break all calls to `gevent.wait` that include Events or Semaphores for `count > 1`, but this a) seems very backward-incompatible, and b) would require classifying all `AbstractLinkables` into resettable and non-resettable

* adds a test that `WaitIterators` can be iterated using just `next()` and without `iter()`, and makes it pass.
* fixes a memory leak alluded to by a comment in the `WaitIterator` code

I did this by decythonizing the WaitIterator class. The pro is that we can now use the Python `__del__`, which cython objects apparently cannot do. What are the cons?

One alternative I just thought of would be to split WaitIterator into a cython `WaitIterator` (no timeout functionality and therefore no possibility of leakage) and a Python subclass `WaitIteratorWithTimeout`.

* makes WaitIterator only return unique objects

Right now a WaitIterator can return the same object many times, which I think is very unexpected and probably pathological. I'm curious for feedback on backward compatibility, but my thought is that the current behavior is so broken (pardon me) that nobody would be relying on it.

PS: I think it might be worthwhile to expose WaitIterator (or something like it) as a documented interface, with `add` and `remove` methods so that users can dynamically choose greenlets to watch.

PPS: I had a lot of fun with the test suite while composing these changes. I especially liked the part where it kept track of reference counts and uncollectible garbage for me. Once I got the tests passing (and not a moment before), I was delighted to have such a detailed test suite.

### DIFF
--- a/src/gevent/_hub_primitives.py
+++ b/src/gevent/_hub_primitives.py
@@ -221,6 +221,10 @@ def wait_on_objects(objects=None, timeout=None, count=None):
     ready. (For example, if count is ``1`` then the function exits
     when any object in the list is ready).
 
+    When ``count`` is ``None`` or ``count`` > 1, some kinds of objects (e.g
+    Event, Semaphore) may lose their readiness before this function returns. To
+    avoid this problem, use :func:`iwait` instead.
+
     If ``timeout`` is provided, it specifies the maximum number of
     seconds ``wait()`` will block.
 
--- a/src/gevent/_hub_primitives.py
+++ b/src/gevent/_hub_primitives.py
@@ -107,6 +107,8 @@ class _WaitIterator(object):
         )
         self._begun = False
 
+        # XXX: If iteration doesn't actually happen, we
+        # could leave these links around!
         self._objects = set(objects)
         for obj in self._objects:
             obj.rawlink(self._switch)
--- a/src/gevent/_hub_primitives.py
+++ b/src/gevent/_hub_primitives.py
@@ -101,48 +101,47 @@ class _WaitIterator(object):
         self._hub = hub
         self._waiter = MultipleWaiter(hub) # pylint:disable=undefined-variable
         self._switch = self._waiter.switch
-        self._timeout = timeout
-        self._objects = objects
-
-        self._timer = None
+        self._timer = (
+            self._hub.loop.timer(timeout, priority=-1)
+            if timeout is not None else None
+        )
         self._begun = False
 
+        self._objects = set(objects)
+        for obj in self._objects:
+            obj.rawlink(self._switch)
+        self._count = count or len(objects)
 
-        # Even if we're only going to return 1 object,
-        # we must still rawlink() *all* of them, so that no
-        # matter which one finishes first we find it.
-        self._count = len(objects) if count is None else min(count, len(objects))
+    def _begin(self):
+        if self._begun:
+            return
 
+        self._begun = True
+        if self._timer is not None:
+            self._timer.start(self._switch, self)
 
     def __iter__(self):
-        # When we begin iterating, we begin the timer.
-        # XXX: If iteration doesn't actually happen, we
-        # could leave these links around!
-        if not self._begun:
-            self._begun = True
-
-            for obj in self._objects:
-                obj.rawlink(self._switch)
-
-            if self._timeout is not None:
-                self._timer = self._hub.loop.timer(self._timeout, priority=-1)
-                self._timer.start(self._switch, self)
         return self
 
     def __next__(self):
-        if self._count == 0:
+        self._begin()
+
+        if not self._objects or self._count == 0:
             # Exhausted
             self._cleanup()
-            raise StopIteration()
+            raise StopIteration
 
-        self._count -= 1
         try:
             item = self._waiter.get()
             self._waiter.clear()
             if item is self:
                 # Timer expired, no more
                 self._cleanup()
-                raise StopIteration()
+                raise StopIteration
+
+            self._count -= 1
+            self._objects.remove(item)
+            self._unlink(item)
             return item
         except:
             self._cleanup()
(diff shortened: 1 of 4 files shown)


jamadden:
Thanks for the PR! I appreciate the effort put into this.

However, gevent does not use finalizers/`__del__`, for many of the same reasons that relying on open files to close themselves issues ` ResourceWarning: unclosed file`, including:

-  They provide non-deterministic behaviour, running code at some arbitrary point in the future
- Such behaviour differs wildly on different implementations of Python (reference counting vs full GC)
- On some versions of CPython, objects with a `__del__` are not eligible for cyclic garbage collection at all

hashbrowncipher:
Ok, I've rebased the `__del__` commits out of the branch.

src/gevent/_hub_primitives.py
```diff
@@ -101,69 +101,73 @@ def __init__(self, objects, hub, timeout, count):
...
-        self._count = len(objects) if count is None else min(count, len(objects))
-
-
-    def __iter__(self):
-        # When we begin iterating, we begin the timer.
         # XXX: If iteration doesn't actually happen, we
         # could leave these links around!
-        if not self._begun:
-            self._begun = True
+        self._objects = set(objects)
```
jamadden: I don't think we can count on objects being hashable. Overall the simpler PR in #1288 is probably better.

hashbrowncipher:
@jamadden now that #1292 and #1288 are merged, I'm curious to hear more about why we should not enforce uniqueness of objects returned by `iwait()`.

In #1292, you mentioned "Unlinking objects as they become ready breaks the use case that deliberately resets them and lets them become ready, up to count number of times (think of queue processors)." It sounds like you're thinking of a use-case like:

```
    def test_iwait_multiple(self):
        sem1 = Semaphore(0)
        sem2 = Semaphore(0)

        def release():
            while True:
                sem1.release()
                gevent.idle()

        let = gevent.spawn(release)
        count = 0
        with gevent.iwait((sem1, sem2), count=20):
            sem1.acquire()
            count += 1

        self.assertEqual(count, 20)
```

But to make this work would require both a) the user knows the number of iterations they expect, in advance of calling `iwait()` and b) the number of iterations is smaller than the number of objects (because WaitIterator automatically reduces any `count` greater than `len(objects)` to `len(objects)`. So I'm having trouble seeing how a use like a queue processor could benefit from this behavior, and the above test-case fails against the code now in master.

You've also mentioned that we shouldn't require that objects passed to iwait be hashable. All of the objects which I can think people might use in this context are hashable, including Popen, Semaphore, Event, greenlet and Greenlet. And anyone who makes their own class implementing `rawlink()` will automatically get hashability based on the object's `id()`, right?

I'm thinking back to my postscript in the PR description. If it isn't possible to modify `iwait()` to only return unique objects, would it be worthwhile to introduce a new interface, where users can add and remove unique objects to be waited upon simultaneously?

hashbrowncipher:
Oops, the test case should be:

```
    def test_iwait_multiple(self):
        sem1 = Semaphore(0)
        sem2 = Semaphore(0)

        def release():
            while True:
                sem1.release()
                gevent.idle()

        let = gevent.spawn(release)
        count = 0
        with gevent.iwait((sem1, sem2), count=20) as iterator:
            for i in iterator:
                i.acquire()
                count += 1

        self.assertEqual(count, 20)
        let.kill()
```

And it fails with: `AssertionError: 2 != 20`

jamadden:
> I'm curious to hear more about why we should not enforce uniqueness of objects returned by iwait().
> You've also mentioned that we shouldn't require that objects passed to iwait be hashable.

In a nutshell, those *new* restrictions are potentially breaking changes, and I don't really see much benefit to them that can't be accomplished today. E.g., if users want to pass in a `set` of objects, they can already do that. Additionally, we'd be adding overhead for the common case of a static list of objects that the user knows by inspection are already unique.

One thing I've learned repeatedly maintaining gevent and other projects is that if something is allowed, somebody out there is probably taking advantage of it.  Whether or not it's explicitly documented. So making changes that tighten contracts, implicitly or explicitly, is a careful balancing act between a lot of factors. (One quick semi-related example: people sometimes pass what are supposed to be positional parameters as keyword arguments, even to internal functions; changing a positional parameter name can wind up breaking code. There's a whole part of the test suite that makes sure that positional parameter names match the standard lib because of that.)

> And anyone who makes their own class implementing `rawlink()` will automatically get hashability based on the object's id(), right

Not if they implement `__eq__` on Python 3; they must explicitly also implement `__hash__`. Otherwise you get a `TypeError`

>  If it isn't possible to modify iwait() to only return unique objects, would it be worthwhile to introduce a new interface, where users can add and remove unique objects to be waited upon simultaneously?

I think the primitive building blocks for that are already public and that could be done today if it was really needed. To consider extending gevent's interface in that direction I'd need to see real use-cases, ideally in different codebases, to suggest that it's a common enough need that can't be easily met otherwise.

I can imagine that there might be a use-case for optionally unlinking objects as they are yielded, but that seems like it's rare, and if that's a problem it might be that the design is too complicated (i.e., i-waiting on and setting the same event from multiple greenlets at the same time is some sort of complicated multi-fan-in-fan-out thing that can probably be simplified). But I haven't thought too much about it.

hashbrowncipher:
Thanks for the detailed explanation. I'd like to clarify two points and suggest one idea (or maybe two).

> Additionally, we'd be adding overhead for the common case of a static list of objects that the user knows by inspection are already unique.

The problem raised in this PR is not related to the _input_ of `iwait`, but rather its _output_. The user may know plenty about their input, but the goal of this PR was to ensure that `iwait` produces sane output by not allowing one (e.g.) Semaphore to be emitted 20 times, while the other 19 items are emitted not at all.

> Not if they implement __eq__ on Python 3; they must explicitly also implement __hash__. Otherwise you get a `TypeError`

Ooh, I didn't know that. Thanks for informing me. As a workaround (and assuming we wanted to go ahead with any version of this PR), we could potentially use a `dict()` indexed by `id(obj)`

Thanks again for your responses. If any of this is news for you, I'd be happy to continue iterating on this PR. If otherwise, it's probably time to close discussion.

jamadden:
Thanks for a great discussion! It's not clear to me right now that these changes are something gevent should move forward with, though.

PR #1292, opened 2018-10-16
Provide a context manager API to prevent memory leaks in gevent.iwait()

hashbrowncipher:
This is a different approach to the `__del__` one proposed in #1290. Just like with open file descriptors, this PR provides (but does not require consumers to use) a context manager API to manage cleanup of WaitIterators.

### DIFF
--- a/src/gevent/_hub_primitives.py
+++ b/src/gevent/_hub_primitives.py
@@ -201,6 +201,8 @@ def iwait_on_objects(objects, timeout=None, count=None):
     .. versionchanged:: 1.1a2
        No longer raise :exc:`LoopExit` if our caller switches greenlets
        in between items yielded by this function.
+    .. versionchanged:: 1.4
+       Add support to use the returned object as a context manager.
     """
     # QQQ would be nice to support iterable here that can be generated slowly (why?)
     hub = get_hub()
--- a/src/gevent/_hub_primitives.py
+++ b/src/gevent/_hub_primitives.py
@@ -168,6 +168,12 @@ class _WaitIterator(object):
                 except: # pylint:disable=bare-except
                     traceback.print_exc()
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, typ, value, tb):
+        self._cleanup()
+
 
 def iwait_on_objects(objects, timeout=None, count=None):
     """
--- a/src/gevent/_hub_primitives.py
+++ b/src/gevent/_hub_primitives.py
@@ -180,6 +180,11 @@ def iwait_on_objects(objects, timeout=None, count=None):
     Iteratively yield *objects* as they are ready, until all (or *count*) are ready
     or *timeout* expired.
 
+    This function allocates resources which must be cleaned up. Consuming the
+    iterator until it is exhausted will automatically clean them up, and it is
+    also possible to use the returned object as a context manager to ensure
+    cleanup occurs.
+
     :param objects: A sequence (supporting :func:`len`) containing objects
         implementing the wait protocol (rawlink() and unlink()).
     :keyword int count: If not `None`, then a number specifying the maximum number
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,12 @@
   Previously such a construction would hang. See :pr:`1288`, provided by Josh
   Snyder.
 
+- There is new documentation on the possibility that `gevent.iwait` can produce
+  memory leaks when not fully consumed. To prevent such situations,
+  `gevent.iwait` now provides optional contextmanager support, which will ensure
+  that all resources are cleaned up. See :pr:`1290`, provided by Josh Snyder.
+
+
 1.3.7 (2018-10-12)
 ==================
 
--- a/src/gevent/__hub_primitives.pxd
+++ b/src/gevent/__hub_primitives.pxd
@@ -58,6 +58,10 @@ cdef class _WaitIterator:
     cdef _begin(self)
     cdef _cleanup(self)
 
+    cpdef __enter__(self)
+    cpdef __exit__(self, typ, value, tb)
+
+
 cpdef iwait_on_objects(objects, timeout=*, count=*)
 cpdef wait_on_objects(objects=*, timeout=*, count=*)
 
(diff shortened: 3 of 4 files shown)


src/gevent/_hub_primitives.py
```diff
@@ -101,30 +101,27 @@ def __init__(self, objects, hub, timeout, count):
...
         self._objects = objects
 
-        self._timer = None
         self._begun = False
 
-        # Even if we're only going to return 1 object,
-        # we must still rawlink() *all* of them, so that no
-        # matter which one finishes first we find it.
-        self._count = len(objects) if count is None else min(count, len(objects))
+        self._objects = set(objects)
```
jamadden: We can't count on objects to be hashable.

src/gevent/_hub_primitives.py
```diff
@@ -101,30 +101,27 @@ def __init__(self, objects, hub, timeout, count):
...
+        self._count = count or len(objects)
```
jamadden: Please return to the old, explicit, form.

src/gevent/_hub_primitives.py
```diff
@@ -101,30 +101,27 @@ def __init__(self, objects, hub, timeout, count):
         self._hub = hub
         self._waiter = MultipleWaiter(hub) # pylint:disable=undefined-variable
         self._switch = self._waiter.switch
-        self._timeout = timeout
+        self._timer = (
```
jamadden: Please return this to the old form. If this object is never iterated, we don't want to allocate a native timer.

src/gevent/_hub_primitives.py
```diff
@@ -133,40 +130,49 @@ def __iter__(self):
     def __next__(self):
         self._begin()
 
-        if self._count == 0:
+        if not self._objects or self._count == 0:
```
jamadden: We can't hash the objects and we can't remove them. This needs to go back to the way it was.

src/gevent/_hub_primitives.py
```diff
@@ -133,40 +130,49 @@ def __iter__(self):
...
             if item is self:
                 # Timer expired, no more
-                self._cleanup()
-                raise StopIteration()
+                self.__exit__(None, None, None)
+                raise StopIteration
+
+            self._count -= 1
+            self._objects.remove(item)
+            self._unlink(item)
```
jamadden: Unlinking objects as they become ready breaks the use case that *deliberately* resets them and lets them become ready, up to *count* number of times (think of queue processors). Please restore.

src/gevent/_hub_primitives.py
```diff
@@ -101,30 +101,27 @@ def __init__(self, objects, hub, timeout, count):
...
+        self._objects = set(objects)
+        self._count = count or len(objects)
 
     def _begin(self):
         if self._begun:
             return
 
         self._begun = True
 
-        # XXX: If iteration doesn't actually happen, we
```
jamadden: This is still true, with the added proviso that the context manager is also not used. Please restore.

src/gevent/_hub_primitives.py
```diff
@@ -235,7 +246,9 @@ def wait_on_objects(objects=None, timeout=None, count=None):
     if objects is None:
         hub = get_hub()
         return hub.join(timeout=timeout) # pylint:disable=
-    return list(iwait_on_objects(objects, timeout, count))
+
+    with iwait_on_objects(objects, timeout, count) as iterator:
```
jamadden: Why do we need this change? `list` will either consume the iterator, or an exception will be raised and the iterator cleans itself up.

src/gevent/_hub_primitives.py
```diff
@@ -101,30 +101,27 @@ def __init__(self, objects, hub, timeout, count):
...
+        self._timer = (
+            hub.loop.timer(timeout, priority=-1) if
+            timeout is not None else None
+        )
         self._objects = objects
 
-        self._timer = None
         self._begun = False
 
-        # Even if we're only going to return 1 object,
```
jamadden: Please restore the comments.

src/gevent/_hub_primitives.py
```diff
@@ -133,47 +130,61 @@ def __iter__(self):
...
+                self.__exit__(None, None, None)
+                raise StopIteration
+
+            self._count -= 1
+            self._objects.remove(item)
+            self._unlink(item)
             return item
         except:
-            self._cleanup()
+            self.__exit__(None, None, None)
```
jamadden: Let's try not to call dunder methods directly. Leave `_cleanup` as it was, call it here, and call it from `__exit__`.

hashbrowncipher:
My apologies for leaving the set/uniqueness changes in this PR. I had meant to leave that discussion to happen in #1290, but carelessness on my part left them in. I'll make the changes you've requested and then I'll ask you my questions about uniqueness in the other PR.

src/gevent/_hub_primitives.py
hashbrowncipher: Good point.

hashbrowncipher:
Ok, I've now made the changes you've requested. Apologies again for including unrelated stuff in this PR.

src/gevent/_hub_primitives.py
```diff
@@ -168,12 +168,23 @@ def _cleanup(self):
...
+    def __exit__(self, typ, value, tb):
+        self._cleanup()
+
 
 def iwait_on_objects(objects, timeout=None, count=None):
     """
     Iteratively yield *objects* as they are ready, until all (or *count*) are ready
     or *timeout* expired.
 
+    This function allocates resources which must be cleaned up. Consuming the
```
jamadden: Could we please get a `.. versionchanged` directive added at the end?

jamadden:
No worries! Looking good now. Let's see what CI says.

src/gevent/_hub_primitives.py
hashbrowncipher: Done!

jamadden:
Thanks! Merged in https://github.com/gevent/gevent/commit/b23342ce1330424e0e0fd2306e64f5e268884030

PR #1311, opened 2018-11-12
Make gevent.socket.getaddrinfo return enums on Python 3

### LINKED ISSUE #1310: Patching socket in Python3 give wrong return values for socket.getaddrinfo
* gevent version: 1.3.7 (installed through pip)
* Python version: 2.7.13 vs 3.5.3
* Operating System: Raspbian GNU/Linux 9 (4.14.79-v7+ #1159 SMP Sun Nov 4 17:50:20 GMT 2018 armv7l GNU/Linux)

### Description:
In python3 it looks like the socket variables are changed.
After running **monkey.patch_socket()** the function **socket.getaddrinfo** does return different/numeric values.

By default socket.getaddrinfo returns values of:
<AddressFamily.AF_INET6: 10> and <SocketKind.SOCK_STREAM: 1>
https://docs.python.org/3/library/socket.html#socket.getaddrinfo

After monky.patch_socket those are just integers 10 and 1

This will give an assert error when you check if the output Socket stream is the same value as socket.SOCK_STREAM

On python2 it is not an issue, because there it is all integers

```python
  File "t.py", line 34, in <module>
    assert gai[0][1] is socket.SOCK_STREAM
AssertionError
```

### What I've run:

```python
# Error on https://github.com/miccoli/pyownet/blob/dev/src/pyownet/protocol.py#L724 when monkey patch is enabled....?
# https://docs.python.org/3/library/socket.html#socket.getaddrinfo

import socket
import sys
from gevent import monkey

print('Monkey patch socket bug in Python 3.5+ ?')
print('Currently running python version: {}'.format(sys.version_info))
print('BEFORE monkey.patch_socket()')
print('Test socket stream socket.SOCK_STREAM')
print(socket.SOCK_STREAM)
print('Stream socket nr: {}'.format(socket.SOCK_STREAM))

print('Get getaddrinfo data for example.com on port 80')
gai = socket.getaddrinfo('example.com', 80, 0, socket.SOCK_STREAM,socket.IPPROTO_TCP)
print(gai)
print('Check if input socket equals with output type value: socket.SOCK_STREAM is gai[0][1]: {}'.format(socket.SOCK_STREAM is gai[0][1]))
assert gai[0][1] is socket.SOCK_STREAM

monkey.patch_socket()

print('AFTER monkey.patch_socket()')
print('Test socket stream socket.SOCK_STREAM')
print(socket.SOCK_STREAM)
print('Stream socket nr: {}'.format(socket.SOCK_STREAM))

print('Get getaddrinfo data for example.com on port 80')
gai = socket.getaddrinfo('example.com', 80, 0, socket.SOCK_STREAM,socket.IPPROTO_TCP)
print(gai)
print('Check if input socket equals with output type value: socket.SOCK_STREAM is gai[0][1]: {}'.format(socket.SOCK_STREAM is gai[0][1]))

print('This will cause an error with the test: assert gai[0][1] is socket.SOCK_STREAM')
assert gai[0][1] is socket.SOCK_STREAM

```

### Output python3:
```
Monkey patch socket bug in Python 3.5+ ?
Currently running python version: sys.version_info(major=3, minor=5, micro=3, releaselevel='final', serial=0)
BEFORE monkey.patch_socket()
Test socket stream socket.SOCK_STREAM
SocketKind.SOCK_STREAM
Stream socket nr: 1
Get getaddrinfo data for example.com on port 80
[(<AddressFamily.AF_INET6: 10>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('2606:2800:220:1:248:1893:25c8:1946', 80, 0, 0)), (<AddressFamily.AF_INET: 2>, <SocketKind.SOCK_STREAM: 1>, 6, '', ('93.184.216.34', 80))]
Check if input socket equals with output type value: socket.SOCK_STREAM is gai[0][1]: True
AFTER monkey.patch_socket()
Test socket stream socket.SOCK_STREAM
SocketKind.SOCK_STREAM
Stream socket nr: 1
Get getaddrinfo data for example.com on port 80
[(10, 1, 6, '', ('2606:2800:220:1:248:1893:25c8:1946', 80, 0, 0)), (2, 1, 6, '', ('93.184.216.34', 80))]
Check if input socket equals with output type value: socket.SOCK_STREAM is gai[0][1]: False
This will cause an error with the test: assert gai[0][1] is socket.SOCK_STREAM
  File "t.py", line 34, in <module>
    assert gai[0][1] is socket.SOCK_STREAM
AssertionError

```

### Output python2:
```
Monkey patch socket bug in Python 3.5+ ?
Currently running python version: sys.version_info(major=2, minor=7, micro=13, releaselevel='final', serial=0)
BEFORE monkey.patch_socket()
Test socket stream socket.SOCK_STREAM
1
Stream socket nr: 1
Get getaddrinfo data for example.com on port 80
[(10, 1, 6, '', ('2606:2800:220:1:248:1893:25c8:1946', 80, 0, 0)), (2, 1, 6, '', ('93.184.216.34', 80))]
Check if input socket equals with output type value: socket.SOCK_STREAM is gai[0][1]: True
AFTER monkey.patch_socket()
Test socket stream socket.SOCK_STREAM
1
Stream socket nr: 1
Get getaddrinfo data for example.com on port 80
[(10, 1, 6, '', ('2606:2800:220:1:248:1893:25c8:1946', 80, 0, 0)), (2, 1, 6, '', ('93.184.216.34', 80))]
Check if input socket equals with output type value: socket.SOCK_STREAM is gai[0][1]: True
This will cause an error with the test: assert gai[0][1] is socket.SOCK_STREAM
```

jamadden:
Fixes #1310

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -46,6 +46,10 @@
   reusable. In particular, the tests are now run with ``python -m
   gevent.tests``. See :issue:`1293`.
 
+- Make a monkey-patched ``socket.getaddrinfo`` return socket module
+  enums instead of plain integers for the socket type and address
+  family on Python 3. See :issue:`1310` reported by TheYOSH.
+
 1.3.7 (2018-10-12)
 ==================
 
--- a/src/gevent/_socketcommon.py
+++ b/src/gevent/_socketcommon.py
@@ -225,7 +225,7 @@ if PY3:
     def _intenum_converter(value, enum_klass):
         try:
             return enum_klass(value)
-        except ValueError:
+        except ValueError: # pragma: no cover
             return value
 
 
--- a/src/gevent/_socketcommon.py
+++ b/src/gevent/_socketcommon.py
@@ -204,10 +204,29 @@ if PY3:
     # the deprecation warning.
     d = getaddrinfo.__doc__
 
-    def getaddrinfo(host, port, family=0, type=0, proto=0, flags=0): # pylint:disable=function-redefined
-        return get_hub().resolver.getaddrinfo(host, port, family, type, proto, flags)
+    def getaddrinfo(host, port, family=0, type=0, proto=0, flags=0):
+        # pylint:disable=function-redefined, undefined-variable
+        # Also, on Python 3, we need to translate into the special enums.
+        # Our lower-level resolvers, including the thread and blocking, which use _socket,
+        # function simply with integers.
+        addrlist = get_hub().resolver.getaddrinfo(host, port, family, type, proto, flags)
+        result = [
+            (_intenum_converter(af, AddressFamily),
+             _intenum_converter(socktype, SocketKind),
+             proto, canonname, sa)
+            for af, socktype, proto, canonname, sa
+            in addrlist
+        ]
+        return result
+
     getaddrinfo.__doc__ = d
     del d
+
+    def _intenum_converter(value, enum_klass):
+        try:
+            return enum_klass(value)
+        except ValueError:
+            return value
 
 
 def gethostbyaddr(ip_address):
--- a/src/gevent/resolver/__init__.py
+++ b/src/gevent/resolver/__init__.py
@@ -4,17 +4,17 @@ from _socket import gaierror
 from _socket import error
 from _socket import getservbyname
 from _socket import getaddrinfo
+from _socket import SOCK_STREAM
+from _socket import SOCK_DGRAM
+from _socket import SOL_TCP
+from _socket import AI_CANONNAME
+from _socket import EAI_SERVICE
+from _socket import AF_INET
+from _socket import AI_PASSIVE
+
 
 from gevent._compat import string_types
 from gevent._compat import integer_types
-
-from gevent.socket import SOCK_STREAM
-from gevent.socket import SOCK_DGRAM
-from gevent.socket import SOL_TCP
-from gevent.socket import AI_CANONNAME
-from gevent.socket import EAI_SERVICE
-from gevent.socket import AF_INET
-from gevent.socket import AI_PASSIVE
 
 # Nothing public here.
 __all__ = []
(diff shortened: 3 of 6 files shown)


minus7:
Unfortunately this seems to break getaddrinfo on Python 3.4, since `SocketKind` does not exist there. Instead it uses `SocketType`:

```
$ python3
Python 3.4.2 (default, Sep 25 2018, 22:02:39) 
[GCC 4.9.2] on linux
>>> import socket
>>> socket.getaddrinfo("localhost", "53")
[(<AddressFamily.AF_INET6: 10>, <SocketType.SOCK_STREAM: 1>, 6, '', ('::1', 53, 0, 0)), ...
```

jamadden:
The `SocketKind` enum [is documented for all Python 3.4](https://docs.python.org/3.4/library/socket.html#constants):

> The AF_* and SOCK_* constants are now `AddressFamily` and `SocketKind` IntEnum collections.
> New in version 3.4.

Unfortunately, in a similar way to [Python 2.2.1](http://python-history.blogspot.com/2013/11/the-history-of-bool-true-and-false.html), that wasn't actually true [until Python 3.4.3](https://github.com/python/cpython/commit/7184bac5446aefcf576bc8a0a666cfd096b86293) and [Python issue 20386](https://bugs.python.org/issue20386). This is not documented in the change notes.

Thank you for bringing this to our attention!

Given that Python 3.4.9 is the current release of Python 3.4, and that Python 3.4.10 is due in March and will end support for Python 3.4, I would encourage you to upgrade your Python version. It's not clear to me that it's reasonable to add special support for three obsolete point releases that are 4+ years old, especially when we don't test with those on CI. Indeed, I'm more inclined to adjust the `python_requires` metadata to specify ` >= 3.4.3`.

minus7:
Raising the minimum version to 3.4.3 sounds reasonable. Note that that will exclude Debian 8 (even including backports) because that only ships 3.4.2. I'll upgrade to Debian 9 then. Thanks

PR #1316, opened 2018-11-14
Python 2: Correct errors from async closed sockets.

jamadden:
Previously we would create a AttributeError, instead of the preferred socket.error(EBADF).

Python 3 already did the right thing.

Discovered because of bad test coverage in test__hub.py that wasn't doing what we thought.

Incremental work on increasing test coverage.


jamadden:
Windows crash is unrelated. See #1317 .

Issue #1318, opened 2018-11-21
Memory leak between io.BufferedWriter and socket.sendall


damz:
* gevent version: 1.3.7
* Python version: cPython 2.7.15
* Operating System: Fedora 29

### Description:

This is a rather weird issue. In the environment above, this leaks:

```python
import gevent.monkey ; gevent.monkey.patch_all()

import io
import socket

import gevent


class SocketFile(object):
    def __init__(self, sock):
        self._sock = sock

    def readable(self):
        return False

    def writable(self):
        return True

    def write(self, data):
        self._sock.sendall(data)
        return len(data)

    @property
    def closed(self):
        return False


one, two = socket.socketpair()

def _read():
    while True:
        buf = one.recv(4096)
        if buf == "":
            break

gevent.spawn(_read)

f = SocketFile(two)
bf = io.BufferedWriter(f)

while True:
    bf.write("test")
    bf.flush()
```

This is because `self._sock.sendall` ends up receiving a `memoryview` object built by `io.BufferedWriter`, and somehow that ends up leaking... but:

 * Replacing `self._sock.sendall(data)` by `self._sock.sendall(data.tobytes())` fixes the leak
 * Just `s.sendall(memoryview("test"))` doesn't seem to leak, so it is not _just_ about passing a `memoryview` to `socket.sendall`

arcivanov:
>  and somehow that ends up leaking
How do you know if it's actually leaking?

> Replacing self._sock.sendall(data) by self._sock.sendall(data.tobytes()) fixes the leak
Also creates a huge allocation completely destroying the point of `memoryview` in the first place.

jamadden:
With this example, I can reproduce an accumulation of `memoryview` objects (a leak) in 2.7.15, but not 3.5+.

In the given example, `bf.flush()` in the loop causes `SocketFile.write` to be called with a different 4-byte memoryview on each iteration. When that gets to `sock.sendall`, it goes to `gevent._socket2._get_memory()`, which does this:
```python
def _get_memory(data):
    try:
        mv = memoryview(data)
        if mv.shape:
            return mv
        # No shape, probably working with a ctypes object,
        # or something else exotic that supports the buffer interface
        return mv.tobytes()
    except TypeError:
        # fixes "python2.7 array.array doesn't support memoryview used in
        # gevent.socket.send" issue
        # (http://code.google.com/p/gevent/issues/detail?id=94)
        return buffer(data)
```

[...]

PR #1319, opened 2018-11-21
Workaround a memoryview leak in py2

damz:
See #1318

### DIFF
--- a/src/gevent/_socket2.py
+++ b/src/gevent/_socket2.py
@@ -70,12 +70,16 @@ else:
 
 def _get_memory(data):
     try:
-        mv = memoryview(data)
-        if mv.shape:
-            return mv
+        if not isinstance(data, memoryview):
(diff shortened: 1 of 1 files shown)


jamadden:
Thanks for the PR. Given the substantial slow down this has for the normal case, I would prefer to find a different solution.

jamadden:
`_get_memory` is in a pure-Python module (gevent._socket2), so you could monkey-patch that function to this version in the interim.

damz:
@jamadden We are not currently affected, we workaround that differently (by removing the `io.BufferedWriter` that was actually completely unnecessary in this case).

jamadden:
> @jamadden We are not currently affected, we workaround that differently (by removing the io.BufferedWriter that was actually completely unnecessary in this case).

Cool, good to know! Thanks again for the report!

PR #1331, opened 2018-12-24
 HTTP/1.0 should support client header: Connection: Keep-Alive

tanchuhan:
HTTP/1.0 should support client header: Connection: Keep-Alive


tanchuhan:
wiki link about Connection header:
https://en.wikipedia.org/wiki/HTTP_persistent_connection#HTTP_1.0
i think we should support Keep-Alive in HTTP/1.0

jamadden:
Thanks for the PR!

The link you provided points out that this is not a part of the standard. Can you explain why gevent should implement a non-standard extension to an obsolete protocol?

jamadden:
Updated and merged in #1490.

Issue #1354, opened 2019-02-20
Monkey patching of existing RLock is broken under CPython3


olliemath:
* gevent version: 1.4.0
* Python version: CPython 3.7.0
* Operating System: Ubuntu 18.04

### Description:

This recently raised issue for Eventlet also applies to Gevent's monkey patching code: https://github.com/eventlet/eventlet/issues/546 That is, under Python3, this code https://github.com/gevent/gevent/blob/master/src/gevent/monkey.py#L485 fails to find and patch existing RLocks.

Here's an example of this:
```python
from __future__ import print_function
import threading
import gevent
from gevent import monkey

lock = threading.RLock()

monkey.patch_all()


def actor(name):
    have_lock = False
    while not have_lock:
        print(name + " trying to get lock")
        have_lock = lock.acquire(blocking=False)
        gevent.sleep()

    print(name + "1")
    gevent.sleep()
    lock.release()

    print(name + "2")


def main():
    ga = gevent.spawn(actor, "A")
    gb = gevent.spawn(actor, "B")
    gevent.sleep()
    ga.join()
    gb.join()

main()
```

Under python2.7 this works as expected with actor B waiting for the lock:
```
A trying to get lock
B trying to get lock
A1
B trying to get lock
A2
B trying to get lock
B1
B2
```
however under python3.7 it does not:
```
A trying to get lock
B trying to get lock
A1
B1
A2
B2
```

The python 3 behaviour is equivalent to the python 2 behaviour without monkey patching.

drpoggi:
I'm not sure what the difference between 2 and 3 is that's causing it to work on one but not the other, but you need to do monkey patching as soon as possible. Try:
```
from __future__ import print_function
from gevent import monkey
monkey.patch_all()
import threading
import gevent

lock = threading.RLock()
```
That works in 3.7.

drpoggi:
Monkey patching an existing RLock isn't supported. The documentation is clear that monkey patching needs to be done first, and if it isn't then unexpected things could happen. The fact that this worked on 2.7 was probably just luck.

jamadden:
There *is* an attempt to find and adjust existing locks at monkey-patch time. Under Python 2, the main motivation for this had to do with the `logging` module; under Python 3, the main motivation is the locks taken during import; since monkey-patching is usually done at import time, this is important. Those have a different implementation and I believe that still works.

[...]

PR #1355, opened 2019-02-20
Use _is_owned method to check RLock ownership.

olliemath:
Fixes PyPY 2.7-7 compatability issue with monkey.patch_all(). Namely, the following raises an attribute error under PyPy2.7-7:
```python
from gevent import monkey
monkey.patch_all()
```

See https://bitbucket.org/pypy/pypy/issues/2962/gevent-cannot-patch-rlock-under-pypy-27-7 for the underlying PyPy change and a minimal dockerfile reproducing the error.

It is possible I can add some backwards compatible patches to the PyPy codebase. But this PR fixes basic monkey patching functionality under PyPy 2.7-7.0.0.

### DIFF
--- a/src/gevent/monkey.py
+++ b/src/gevent/monkey.py
@@ -485,11 +485,10 @@ def _patch_existing_locks(threading):
     gc = __import__('gc')
     for o in gc.get_objects():
         if isinstance(o, rlock_type):
-            if hasattr(o, '_owner'): # Py3
-                if o._owner is not None:
+            if o._is_owned():
+                if hasattr(o, '_owner'): # Py3
                     o._owner = tid
-            else:
-                if o._RLock__owner is not None:
+                else:
                     o._RLock__owner = tid
         elif isinstance(o, _ModuleLock):
             if o.owner is not None:


olliemath:
Related, but different issue with the `_patch_existing_locks` function: https://github.com/gevent/gevent/issues/1354

jamadden:
Thanks for the report. CI is now failing for Python 2 and PyPy2.

olliemath:
As a direct consequence of the patch? Hmm .. is there a way to run the tests locally? I tried
```
pip install ci-requirements.txt
tox
```
but didn't even get past the import stage.

jamadden:
Yes, I believe this is a direct consequence of the patch:
```
| GEVENTTEST_LEAKCHECK=1 GEVENT_RESOLVER=thread GEVENT_RESOLVER_NAMESERVERS=8.8.8.8 /home/travis/.runtimes/snakepit/python2.7.15 -u -mgevent.tests.test__threading_holding_lock_while_monkey
    File "/home/travis/.runtimes/versions/python2.7.15/lib/python2.7/runpy.py", line 174, in _run_module_as_main
      "__main__", fname, loader, pkg_name)
    File "/home/travis/.runtimes/versions/python2.7.15/lib/python2.7/runpy.py", line 72, in _run_code
      exec code in run_globals
    File "/home/travis/build/gevent/gevent/src/gevent/tests/test__threading_holding_lock_while_monkey.py", line 8, in <module>
      monkey.patch_all()
    File "/home/travis/.runtimes/versions/python2.7.15/lib/python2.7/threading.py", line 216, in __exit__
      self.release()
    File "/home/travis/.runtimes/versions/python2.7.15/lib/python2.7/threading.py", line 204, in release
      raise RuntimeError("cannot release un-acquired lock")
  RuntimeError: cannot release un-acquired lock
```

> Hmm .. is there a way to run the tests locally?
> pip install ci-requirements.txt

(Did you leave out the -r?)

tox by itself should work (it creates its own virtual environments), no need to install anything beyond tox. But to debug you'll probably want to use your own virtual environment. AFAICS, the steps outlined at http://www.gevent.org/install.html#development for running tests are still basically the correct ones. If you have a specific issue I may be able to provide a specific solution (and of course a docs patch would be welcome.) (Also notice that the CI runs print the exact command used to execute each test file, so given a development virtual environment with gevent installed, running `python2 -mgevent.tests.test__threading_holding_lock_while_monkey` should reproduce the problem.

olliemath:
Ah, OK `_is_owned` is not working as I thought it would. The code passed my naive tests because there were no acquired locks present.

It seems like there are two choices: push the PyPy team for some compatibility fixes, or give up on the ability to patch existing acquired RLocks in newer versions of PyPy2.

olliemath:
I'll turn this into an issue to keep track of it.

PR #1378, opened 2019-03-27
Update process.py example to work on windows.

riannucci:
The "Unable to run on windows" warning is actually about uname/ls, not about gevent :)


riannucci:
Force push was to fix the commit message

jamadden:
Thanks!

PR #1439, opened 2019-07-25
monkey: allow to execute a module rather than only a script

jd:


jd:
Not doing what I wanted to do in the end — closing for now.

PR #1440, opened 2019-07-25
monkey: allow to execute a module rather than only a script

jd:


src/gevent/monkey.py
```diff
@@ -1057,14 +1057,18 @@ def patch_all(socket=True, dns=True, time=True, select=True, thread=True, os=Tru
 
 
 def main():
+    import runpy
```
jamadden: Please do not move the import. It was purposefully *after* monkey-patching had been done.

src/gevent/monkey.py
```diff
@@ -1111,18 +1113,15 @@ def _get_script_help():
...
+USAGE: ``python -m gevent.monkey [MONKEY OPTIONS] [--module] (script|module) [SCRIPT OPTIONS]``
 
 If no OPTIONS present, monkey patches all the modules it can patch.
-You can exclude a module with --no-module, e.g. --no-thread. You can
-specify a module to patch with --module, e.g. --socket. In the latter
+You can exclude a module with --no-<module>, e.g. --no-thread. You can
+specify a module to patch with --<module>, e.g. --socket. In the latter
 case only the modules specified on the command line will be patched.
 
-.. versionchanged:: 1.3b1
```
jamadden: Please do not remove documentation that remains relevant.

src/gevent/monkey.py
```diff
@@ -1111,18 +1113,15 @@ def _get_script_help():
...
-    The *script* argument can now be any argument that can be passed to `runpy.run_path`,
-    just like the interpreter itself does, for example a package directory containing ``__main__.py``.
-    Previously it had to be the path to
-    a .py source file.
+The default behavior is to execute the script passed as argument. If you with
+to run a module instead, pass the `--module` argument before the module name.
```
jamadden: This needs a `.. versionchanged:: 1.5` block.

jd: Those blocks are cool to generate doc but they are quite ugly to render in the script help in your terminal…

jamadden:

navytux:
For the reference, [gpython](https://pypi.org/project/pygolang/#gpython) ([2](https://lab.nexedi.com/kirr/pygolang/blob/master/gpython/__init__.py)) also mimics interface of standard `python` interpreter while providing execution environment with gevent preactivated. It supports `-c`, `-m`, `<file>`, interactive console, etc...

PR #1442, opened 2019-07-31
Fix #1441 - FileObjectPosix should use 'str' in text mode

### LINKED ISSUE #1441: FileObject with mode "r" reads bytes instead of string
* gevent version: 1.4.0
* Python version: 3.5.1
* Operating System: macOS 10.14.5

### Description:

I am using `gevent.fileobject.FileObject` with mode "r" (read, not binary). In Python 3, I expect this mode to return `str` objects when reading the file content (only passing binary mode will read `bytes`). However, I am getting `bytes` and not `str` with this mode.

With regular Python file object:
```python
>>> fd = open("foo.json", "r")
>>> type(fd.read())
<class 'str'>
```

With `gevent.fileobject.FileObject`:
```python
>>> from gevent.fileobject import FileObject
>>> fd = open("/tmp/bla.json", "r")
>>> fo = FileObject(fd, mode="r")
>>> type(fo.read())
<class 'bytes'>
```

This makes it impossible to use `FileObject` with `json.load` (which reads the content by itself, not giving me a chance to decode the content).

wiggin15:
This PR fixes the different issues with FileObjectPosix mentioned in #1441.
This PR introduces a few changes that break backward-compatibility, but for good reasons:
1. Strictly enforcing `mode` prevents strange modes that are not supposed to be possible but could previously be passed, like `wU` and `rbt`. It also fixes some of the issues mentioned in #1441, like when passing `rUt`. The documentation already explicitly mentioned the supported modes, so any code that may break was already using the class improperly.
2. `translate_newlines` was renamed, but it was only used as a workaround specifically for newlines until now, while the new name, `text_mode`, is more general and better explains its effect. I don't know if it is really used directly outside of the package much.

### DIFF
--- a/src/gevent/_fileobjectcommon.py
+++ b/src/gevent/_fileobjectcommon.py
@@ -60,11 +60,10 @@ class FileObjectBase(object):
     )
 
 
-    # Whether we are translating universal newlines or not.
-    _translate = False
-
-    _translate_encoding = None
-    _translate_errors = None
+    _text_mode = False
+    _text_encoding = None
+    _text_errors = None
+    _text_newline = None
 
     def __init__(self, io, closefd):
         """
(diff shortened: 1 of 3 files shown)


jamadden:

wiggin15:
Hi @jamadden can you please take a look at this?
FileObjectPosix is currently not working with the correct types in Python 3. This issue is stopping us from moving to Python 3 with gevent. We will have to fork gevent if this isn't fixed and we prefer to avoid doing that. Thanks!

PR #1445, opened 2019-08-05
Dynamic class in gevent.resolver.thread.Resolver.__repr__

jparise:
The class name used in `gevent.resolver.thread.Resolver.__repr__` was
hard-coded. This meant that subclasses needed to reimplement their own
`__repr__`; otherwise, they'd appear as this parent class in debugging
output like that from print_run_info().

This also has the nice side effect of fixing a typo (`resolver_thread`
to `resolver.thread`).


jparise:
Test failures are also occurring on HEAD and should be unrelated to this diff.

jamadden:

PR #1449, opened 2019-08-15
Multiple fixes and enhancements for monkey pluggability

### LINKED ISSUE #1447: Monkey pluggability enhancements and bugs
1. `_check_repatching` intentionally deletes `kwargs` that contain `patch_all` settings from external plugins, resulting in loss of state: https://github.com/gevent/gevent/blob/master/src/gevent/monkey.py#L957

2.  Similarly, would be nice need to expose that patch_all state to external plugin through public API.

3. `_patch_module` has two problems (https://github.com/gevent/gevent/blob/master/src/gevent/monkey.py#L362):
      a. `gevent.` is hardcoded.
      b. the construct is only capable of handling one level, i.e. `gevent.a`. If used with `gevent.a.b`, the expression fails to evaluate properly.

4. The current event infrastructure lacks capability to support default patching whenever `patch_all` has been invoked. Event should allow checking not only whether the option was specified but also if option was explicitly disabled, i.e. allow to distinguish between `patch_all()`, `patch_all(myplugin=True)` and `patch_all(myplugin=False)`.

5. In cases where an `item` of a module is being reduced it may be necessary to substitute the `__module__` of a `newitem` during patching to make sure that the pickle encodes the proper module name into its stream, i.e. the name of the module *being* patched, not the name of the module *hosting* a patch.

Patch is incoming.

arcivanov:
1. `_check_repatching` now merges kwargs into state
2. `_get_patch_all_state` allows to retrieve the patch_all configuration state
3.a `_patch_module` now allows to specific `_package_prefix` parameter that defaults to 'gevent.'
3.b `import_module` is now used in `_patch_module` to allow handling of package names of any depth.
4. `patch_item` and both `patch_modules` now accept argument `_patch_module` for item's `__module__` override.

fixes #1447

### DIFF
--- a/src/gevent/monkey.py
+++ b/src/gevent/monkey.py
@@ -119,6 +119,7 @@ For details, see the :func:`main` function.
 from __future__ import absolute_import
 from __future__ import print_function
 import sys
+from importlib import import_module
 
 __all__ = [
     'patch_all',
@@ -255,11 +256,18 @@ def get_original(mod_name, item_name):
 _NONE = object()
 
 
-def patch_item(module, attr, newitem):
+def patch_item(module, attr, newitem, _patch_module=False):
     olditem = getattr(module, attr, _NONE)
     if olditem is not _NONE:
         saved.setdefault(module.__name__, {}).setdefault(attr, olditem)
     setattr(module, attr, newitem)
+    if _patch_module:
+        if olditem is not None and newitem is not None:
+            try:
+                newitem.__module__ = olditem.__module__
+            except (TypeError, AttributeError):
+                # We will let this fail quietly
+                pass
 
 
 def remove_item(module, attr):
@@ -290,7 +298,8 @@ def patch_module(target_module, source_module, items=None,
                  _warnings=None,
                  _notify_will_subscribers=True,
                  _notify_did_subscribers=True,
-                 _call_hooks=True):
+                 _call_hooks=True,
+                 _patch_module=False):
     """
     patch_module(target_module, source_module, items=None)
 
@@ -336,7 +345,8 @@ def patch_module(target_module, source_module, items=None,
         return False
 
     for attr in items:
-        patch_item(target_module, attr, getattr(source_module, attr))
+        patch_item(target_module, attr, getattr(source_module, attr),
+                   _patch_module=_patch_module)
 
     if _call_hooks:
         __call_module_hook(source_module, 'did', target_module, items, _warnings)
@@ -357,11 +367,13 @@ def _patch_module(name,
                   _warnings=None,
                   _notify_will_subscribers=True,
                   _notify_did_subscribers=True,
-                  _call_hooks=True):
+                  _call_hooks=True,
+                  _patch_module=False,
+                  _package_prefix='gevent.'):
 
-    gevent_module = getattr(__import__('gevent.' + name), name)
+    gevent_module = import_module(_package_prefix + name)
     module_name = getattr(gevent_module, '__target__', name)
-    target_module = __import__(module_name)
+    target_module = import_module(module_name)
 
     patch_module(target_module, gevent_module, items=items,
                  _warnings=_warnings,
@@ -386,7 +398,8 @@ def _patch_module(name,
                          _warnings=_warnings,
                          _notify_will_subscribers=False,
                          _notify_did_subscribers=False,
-                         _call_hooks=False)
+                         _call_hooks=False,
+                         _patch_module=_patch_module)
             saved[alternate_name] = saved[module_name]
 
     return gevent_module, target_module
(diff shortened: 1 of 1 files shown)


jamadden:

src/gevent/monkey.py
```diff
@@ -255,11 +256,18 @@ def get_original(mod_name, item_name):
 _NONE = object()
 
 
-def patch_item(module, attr, newitem):
+def patch_item(module, attr, newitem, _patch_module=False):
     olditem = getattr(module, attr, _NONE)
     if olditem is not _NONE:
         saved.setdefault(module.__name__, {}).setdefault(attr, olditem)
     setattr(module, attr, newitem)
+    if _patch_module:
```
jamadden: I'm not sure this is a good idea in general. One example that springs to mind is pickling; another is that it may tend to produce very confusing `repr` results. Why would we want to do this?

If this is something that's for very specialized use-cases, perhaps that would be better served in the hook functions for the module? (See below for some general comments about underscore-prefixed things.)

src/gevent/monkey.py
```diff
@@ -119,6 +119,7 @@ def patch_psycopg(event):
 from __future__ import absolute_import
 from __future__ import print_function
 import sys
+from importlib import import_module
```
jamadden: I was concerned about this extra import either leading to more imports that might cause issue with monkey-patching, or with it taking time for cases that we might not need it. 

But on Python 3, it's imported by default anyway during startup, so those are completely moot.

On Python 2, it's a trivial file with no imports besides `sys`.

src/gevent/monkey.py
```diff
@@ -357,11 +367,13 @@ def _patch_module(name,
                   _warnings=None,
                   _notify_will_subscribers=True,
                   _notify_did_subscribers=True,
-                  _call_hooks=True):
+                  _call_hooks=True,
+                  _patch_module=False,
+                  _package_prefix='gevent.'):
```
jamadden: In the tradition of 'explicit is better than implicit', it would probably be better to make this a required argument instead of defaulting.

Instead of an extra argument, what if `name` was required to be the fully-qualified name instead of the simple module name?  That would allow patches that consist only of a module, not a package. (With a small amount of work, that could generalize to allowing arbitrary objects to be the patch source.)

src/gevent/monkey.py
```diff
@@ -357,11 +367,13 @@ def _patch_module(name,
...
 
-    gevent_module = getattr(__import__('gevent.' + name), name)
+    gevent_module = import_module(_package_prefix + name)
```
jamadden: Probably better named `source_module` at this point.

src/gevent/monkey.py
```diff
@@ -951,9 +964,15 @@ def patch_signal():
     _patch_module("signal")
 
 
+def _get_patch_all_state():
```
jamadden: This function doesn't seem to be used here.

Here's my general philosophy about the underscore prefixed functions and arguments in this module: They're an implementation detail, subject to change. (That's obvious.) But we also know that, as external plugins get going, until we feel ready to publish some sort of official monkey-patching API, they'll be used outside the gevent package. If functionality that gevent itself doesn't use is only used by those external plugins, its much more likely that that functionality is going to break in a way that is not caught by the gevent tests, or be refactored away without planning too. So it's best to keep the implementation details to a minimum and be sure that gevent has a use-case for them all.

In this case, it looks like it would be easy to make `_check_repatching` use this function.

src/gevent/monkey.py
```diff
@@ -951,9 +964,15 @@ def patch_signal():
...
+    key = '_gevent_saved_patch_all'
+    return saved.get(key)
+
+
 def _check_repatching(**module_settings):
     _warnings = []
     key = '_gevent_saved_patch_all'
+    module_settings.update(module_settings['kwargs'])
```
jamadden: It's not clear to me what difference this makes. Could you clarify please? It may be missing a test case, or perhaps that's the reason for the broken test.

jamadden (commented):
Thanks again for this PR. I'm very interested in developing a patching API that's truly usable and useful for external plugins.

gevent master is in a good state again after the SSL updates broke things, so now [on Travis](https://travis-ci.org/gevent/gevent/builds/572391420) we can see one test failure that seems to be related to these changes. I also asked for some clarification and discussion on some of the changes.

src/gevent/monkey.py
jamadden: Oh. Silly me. I just reviewed #1447 and see where the problem was mentioned. So this either is the reason for the test failure, or it definitely needs a specific test.

src/gevent/monkey.py
jamadden: Reviewing #1447 I see that pickling was specifically mentioned as the motivation for this. I'm still not convinced, I'd need to see a specific example to understand what problem this solves that can't be solved by the object itself.

src/gevent/monkey.py
arcivanov: I have no problem with any aesthetic API changes (argument requirement, order, naming, refactoring).

src/gevent/monkey.py
arcivanov: I have no problem with any aesthetic API changes (argument requirement, order, naming, refactoring).

src/gevent/monkey.py
arcivanov: This function is supposed to be used externally by plugins. I was hesitant to make it public, but then again maybe I should've.

I have no problem with any aesthetic API changes (argument requirement, order, naming, refactoring).

jamadden:
As there's been no movement on this PR in seven years and there are now conflicts, I'm going to go ahead and close it.

PR #1454, opened 2019-08-29
Fix references to six.PY3 in case python4 becomes a thing

asottile:
Detected using [flake8-2020](https://github.com/asottile/flake8-2020)


jamadden:
Thanks for the PR! The test failures were unrelated and due to external changes.

PR #1455, opened 2019-09-04
API docs: remove duplicated entries for gevent.{queue,server}

FelixSchwarz:
Trivial change: gevent.queue / gevent.server are listed twice in the API docs.


jamadden:
Heh, I wonder how we managed to do that. Thanks for the fix!

PR #1521, opened 2020-01-27
Don't check to see if event is set before waiting

### LINKED ISSUE #1520: Greenlets that start waiting for an event after it was set awaken before greenlets that started waiting before it was set
* gevent version: 1.5a4.dev0
* Python version: cPython 3.8.0 from [Miniconda](https://docs.conda.io/en/latest/miniconda.html)
* Operating System: Ubuntu 18.10 (Linux 4.19.84-microsoft-standard under [WSL2](https://docs.microsoft.com/en-us/windows/wsl/wsl2-index) on Windows 10 Version 10.0.19041 Build 19041)

### Description:

I've created an object that stores a reusable gevent event. Once the event is set, it's critical that it is cleared *immediately* afterwards - but not before other greenlets that were waiting on it have had a chance to awaken.

1.5a3 [states](https://github.com/gevent/gevent/blob/367047681c79e04f96d645e3676c809f16d2ff98/src/gevent/event.py#L45) that waiting greenlets awaken in the order in which they waited. The event clearing must happen last, so in order to guarantee that, I've queued it up *after* the event has been set.

The clearing event happens first, before anything else that started waiting before it. While I admit that it's somewhat illogical for something to start waiting after its awakening condition is true, this is clearly inconsistent behaviour.

### What I've run:

Here I've defined one instance of a class Signal, which carries a reusable gevent event and some data that's associated with it. Its intent is to allow for broadcasting data between greenlets in my programme.

Because these signals can be reused, it's important that their internal gevent events are cleared as soon as possible.

```python
import gevent
import gevent.event

class Signal:
    def __init__(self):
        self._event = gevent.event.Event()
        self._data = None

    def fire(self, data):
        self._data = data
        self._event.set()
        self.unfire()

    def unfire(self):
        print("Waiting to unfire")
        await_signal()
        print("Unfiring")
        self._event.clear()
        self._data = None

data_signal = Signal()

def emit_signal(data=None):
    data_signal.fire(data)

def await_signal(timeout=None):
    if not data_signal._event.wait(timeout):
        raise TimeoutError
    return data_signal._data

def start_multiple_greenlets():
    gevent.spawn(signal_waiter)
    gevent.spawn(signal_sender)
    gevent.idle()

def signal_sender(data="data string"):
    print("Sending signal with data={}".format(data))
    emit_signal(data=data)

def signal_waiter(timeout=5.0):
    data = await_signal(timeout=timeout)
    print("Signal recieved with data={}".format(data))

for i in range(10):
    start_multiple_greenlets()
```
As of 1.5a3, the output data is as follows:
```
Sending signal with data=data string
Waiting to unfire
Unfiring
Signal recieved with data=None
```
This indicates that the signal was unfired (and the data destroyed) before `signal_waiter` got a chance to access the data.

But `signal_waiter` was the very first thing that was called, and was unambiguously called before `unfire` - and therefore *waited* before `unfire`. Therefore, `signal_waiter` should have awakened before `unfire`.

I can't imagine this is intentional behaviour.

rossjrw:
Should fix #1520.

#1520 details a bug in which a waiter that started waiting on an event after it was set will finish waiting before waiters that started waiting after the event was set. (Try saying that ten times fast.)

I can trace this back to `_wait` in `AbstractLinkable`, which performs a pre-wait check to see if the event is already set:

https://github.com/gevent/gevent/blob/367047681c79e04f96d645e3676c809f16d2ff98/src/gevent/_abstract_linkable.py#L204-L209

This causes it to finish executing before waiters that are actually waiting.

Removing that pre-wait check results in the correct output from the script in #1520:
```
Sending signal with data=data string
Waiting to unfire
Signal recieved with data=data string
Unfiring
```

### DIFF
--- a/src/gevent/_abstract_linkable.py
+++ b/src/gevent/_abstract_linkable.py
@@ -202,9 +202,6 @@ class AbstractLinkable(object):
         return None # pragma: no cover all extent subclasses override
 
     def _wait(self, timeout=None):
-        if self.ready():
-            return self._wait_return_value(False, False)
-
         gotit = self._wait_core(timeout)
         return self._wait_return_value(True, gotit)
 


jamadden:
I don't think this is the right behaviour. This will cause things to needlessly take a trip around the event loop, executing syscalls, allocating memory, etc, when they otherwise wouldn't — in fact, it introduces lots of additional non-determinism that wasn't present before; it may "fix" the case described in #1520, but no doubt it causes other issues. 

I'm actually amazed that this didn't trigger a test failure; I thought there was a test specifically for the existing behaviour, although I don't remember exactly what it might be off the top of my head.

PR #1548, opened 2020-04-03
AIX: fix compilation of libuv

wiggin15:
Added the files and libraries needed for compilation.

The macro _LINUX_SOURCE_COMPAT is needed to prevent duplicate error definitions that result in a long error that says:

    deps/libuv/src/uv-common.c:200:36: error: duplicate case value

AIX defines two errors, ENOTEMPTY and EEXIST, as the same value, unless the macro is set.


jamadden:
Thank you. I don't have any way to verify that so I'll take your word for it 😄

PR #1549, opened 2020-04-03
Solaris 10: define SUNOS_NO_IFADDRS to make libuv compile correctly

wiggin15:
gevent fails to compile on Solaris 10:

    deps/libuv/src/unix/sunos.c:32:22: fatal error: ifaddrs.h: No such file or directory
     # include <ifaddrs.h>

Solaris 10 doesn't have the "ifaddrs" interface. libuv have a flag to support this platform, we just need to turn it on.
See libuv issue here: https://github.com/libuv/libuv/issues/1458
The detection code is similar to what's in `psutil`, see: https://github.com/giampaolo/psutil/blob/master/setup.py#L298

### DIFF
--- a/src/gevent/libuv/_corecffi_build.py
+++ b/src/gevent/libuv/_corecffi_build.py
@@ -7,9 +7,10 @@
 # this file to know what functions are created and available on the generated
 # module.
 from __future__ import absolute_import, print_function
-import sys
 import os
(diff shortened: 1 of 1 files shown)


jamadden:
Thank you. I have no (easy) way to verify so again I'll take your word for it.

PR #1633, opened 2020-05-29
Add support for Python 3.9

jamadden:
Relying on an unreleased version of greenlet, for the moment.

Biggest (nearly only) functional change is supporting user/group/extra_groups/umask in Popen.


radzhome:
When will a version with this update be released?

philihp:
Any update? :pray:

jamadden:
This was also released last year in 20.6.0. The PyPI page shows it, as does [the change log](http://www.gevent.org/changelog.html#id21).
Untitled

philihp:
Cool, so http://www.gevent.org/install.html#supported-platforms needs to be updated?

PR #1648, opened 2020-06-19
Configure libev to use vdso where it is available

hashbrowncipher:
I noticed numerous time-related syscalls in my interpreter's strace output, and I was wondering whether we can feasibly turn on vDSO. It turns out that it works out of the box in a `manylinux2014` environment, but that in `manylinux2010`, we must force linking against `librt`. I built a `manylinux2010` wheel with this configuration and observed that the `clock_gettime` and `gettimeofday` syscalls vanished.


jamadden:
Can you rebase on master, please?

PR #1655, opened 2020-07-03
Use slots for SSLContext for Python 2.7.9+

### LINKED ISSUE #1654: SSLContext not using __slots__ for Python 2.7.9+
* gevent version: master (20.6.2)
* Python version: CPython2.7.9+
* Operating System: Ubuntu 18.04

### Description:

SSLContext uses `__slots__`  in Python3 but not Python2.7.9+. A SSLContext instance uses 416B of memory in Python 2.7.9+.

Script:
```
from gevent import ssl as gssl
import ssl
from pympler.asizeof import asizeof
print(asizeof(ssl.SSLContext(ssl.PROTOCOL_SSLv23)))
print(asizeof(gssl.SSLContext(ssl.PROTOCOL_SSLv23)))
```

Output:
```
128
416
```

carsonip:
Memory usage of SSLContext goes down from 416B to 128B after patching.

Fixes #1654

### DIFF
--- a/src/gevent/_sslgte279.py
+++ b/src/gevent/_sslgte279.py
@@ -74,6 +74,9 @@ class _fileobject(getattr(__ssl__, '_fileobject', object)): # pylint:disable=no-
 orig_SSLContext = __ssl__.SSLContext # pylint: disable=no-member
 
 class SSLContext(orig_SSLContext):
+
+    __slots__ = ()
+
     def wrap_socket(self, sock, server_side=False,
                     do_handshake_on_connect=True,
                     suppress_ragged_eofs=True,


jamadden:
Interesting. I take it these 300 bytes are a major concern for some scenario? Could you please describe that scenario?

carsonip:
I am not bothered by the 300B. Just that I see slots usage all over the place and gevent Python 3 ssl module already has it, and I wonder why it is not used in sslgte279. So far I don't see any drawbacks in terms of performance or compatibility since CPython SSLContext does not have an instance dict anyway. I assume that every SSLSocket has a context and some Python 2 users may like the 300B memory savings. Please let me know your views on this. Thanks!

jamadden:
Python 3's SSLContext does have a `__dict__`, both gevent's version and the stdlib's version (from 3.7 and on).
```pycon
Python 3.9.0b3 (default, Jun 13 2020, 08:23:25)
[Clang 11.0.3 (clang-1103.0.32.29)] on darwin
>>> import gevent.ssl
>>> gevent.ssl.SSLContext()
<gevent._ssl3.SSLContext object at 0x101c10b90>
>>> gevent.ssl.SSLContext().__dict__
{}

>>> import ssl
>>> ssl.SSLContext()
<ssl.SSLContext object at 0x1020eab90>
>>> ssl.SSLContext().__dict__
{}
```

When there's a slight behaviour difference between Python 2 and Python 3, or 3.x and 3.x+1, unless it's a compatibility issue, gevent tends to try to lean towards the newer behaviour (increased functionality and/or decreased strictness). Why not bring the benefits of newer stdlibs to all gevent consumers?

It's possible there are consumers in the wild relying on SSLContext having an instance dictionary. That code would work on Python 3 (stdlib or gevent) or Python 2 (gevent), giving an easy path between the versions.

Given the above, unless there's a compelling reason, this is a change I would be very hesitant to make.

carsonip:
@jamadden Thanks for explaining your thoughts. I saw the `__slots__ = ()` in _ssl3 and thought that Python3 SSLContext doesn't have instance dict without considering the base class definition. I understand and agree that there is no compelling reason to use slots in SSLContext. I am closing the issue and PR now.

Issue #1674, opened 2020-09-22
gevent==20.6.2 should require a specific version of greenlets


dlamblin:
* gevent version: 20.6.2
* greenlets version 0.4.17
* Python version: cPython 3.7.0 downloaded from docker python:3.7.0-slim-stretch
* Operating System: debian stretch

### Description:

My Apache Airflow FlaskAppBuilder webserver installed and ran fine 2h ago with greenlets version v0.4.16;
However making a config change to my dockerfile triggered a rebuild with reinstall which didn't pin the greenlet and brought in the 50 minute old v0.4.17. This logged a complaint about binary incompatibility and didn't serve any pages.

```
/usr/local/lib/python3.7/importlib/_bootstrap.py:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
  return f(*args, **kwds)
[2020-09-22 11:43:24 +0000] [628] [INFO] Starting gunicorn 20.0.4
[2020-09-22 11:43:24 +0000] [628] [INFO] Listening at: http://0.0.0.0:8443 (628)
[2020-09-22 11:43:24 +0000] [628] [INFO] Using worker: gevent
[2020-09-22 11:43:24 +0000] [631] [INFO] Booting worker with pid: 631
[2020-09-22 11:43:24 +0000] [632] [INFO] Booting worker with pid: 632
/usr/local/lib/python3.7/importlib/_bootstrap.py:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
  return f(*args, **kwds)
[2020-09-22 11:43:24 +0000] [633] [INFO] Booting worker with pid: 633
/usr/local/lib/python3.7/site-packages/gunicorn/workers/ggevent.py:53: MonkeyPatchWarning: Monkey-patching ssl after ssl has already been imported may lead to errors, including RecursionError on Python 3.6. It may also silently lead to incorrect behaviour on Python 3.7. Please monkey-patch earlier. See https://github.com/gevent/gevent/issues/1016. Modules that had direct imports (NOT patched): ['urllib3.util.ssl_ (/usr/local/lib/python3.7/site-packages/urllib3/util/ssl_.py)', 'urllib3.util (/usr/local/lib/python3.7/site-packages/urllib3/util/__init__.py)'].
  monkey.patch_all()
```

Rebuilding with 0.4.16 and everything is fine again.

[...]

PR #1676, opened 2020-09-22
Support and require greenlet 0.4.17

jamadden:
This provides native handling of contextvars on Python 3.7+, which we now use.

(Note to self: The upcoming release of PyPy3.7 will probably require revisiting this and either applying the monkey-patches there, or requesting a change to their greenlet implementation to match.)

This incidentally includes (test) fixes for versions of Python that have  https://bugs.python.org/issue31122 fixed, including 3.9 and a backport to a forthcoming 3.8.

The change to CHANGES.rst was unintended, I'll revert.

### DIFF
--- a/scripts/releases/make-manylinux
+++ b/scripts/releases/make-manylinux
@@ -72,9 +72,9 @@ if [ -d /gevent -a -d /opt/python ]; then
         # The downside is that we must install dependencies manually.
         # NOTE: We can't upgrade ``wheel`` because ``auditwheel`` depends on
         # it, and auditwheel is installed in one of these environments.
-        python -mpip install -U "cython >= 3.0a5" cffi greenlet setuptools
+        python -mpip install -U "cython >= 3.0a5" cffi 'greenlet >= 0.4.17' setuptools
         time (python setup.py bdist_wheel -q > /dev/null)
-        PATH="$OPATH" auditwheel repair --plat manylinux2010_x86_64 dist/gevent*.whl
+        PATH="$OPATH" auditwheel repair dist/gevent*.whl
         cp wheelhouse/gevent*.whl /gevent/wheelhouse
 
         python -mpip install -U --no-compile `ls dist/gevent*whl`[test]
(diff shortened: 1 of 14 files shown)


scripts/releases/make-manylinux
```diff
@@ -72,9 +72,9 @@ if [ -d /gevent -a -d /opt/python ]; then
         # The downside is that we must install dependencies manually.
         # NOTE: We can't upgrade ``wheel`` because ``auditwheel`` depends on
         # it, and auditwheel is installed in one of these environments.
-        python -mpip install -U "cython >= 3.0a5" cffi greenlet setuptools
+        python -mpip install -U "cython >= 3.0a5" cffi 'greenlet >= 0.4.17' setuptools
```
IrinaAlexandra, line 75: What do you mean 'yanked' exactly? I am asking because we had an issue in production, same as the one [here](https://github.com/gevent/gevent/issues/1674) and I am trying to get a better understanding of what happened.

jamadden: 'yanked' is a feature on PyPI that prevents a release from being automatically discovered. It's still there but you have to ask for it exactly. The greenlet version was published, then when this issue was discovered it was yanked. When gevent 20.9 was released, it was unyanked.

IrinaAlexandra: oh, that makes sense. Thank you very much kind sir :smile:

davidism:
Did you follow up to confirm that contextvars work with gevent on PyPy? It seems they still haven't updated their version of greenlet. https://foss.heptapod.net/pypy/pypy/-/issues/3408

jamadden:
gevent does not currently test or support PyPy 3.7, so it doesn't matter.

davidism:
PyPy 3.7 7.3.7 released last week and updated greenlet, so it should be ready when gevent supports that version.

jamadden:
Thanks! Bumping the supported PyPy version is high on my priority list, right after greenlet 2.0.

PR #1689, opened 2020-10-21
Set to dnf

fabaff:
`dnf` is the default CLI tool for dealing with packages nowadays.

### DIFF
--- a/docs/development/installing_from_source.rst
+++ b/docs/development/installing_from_source.rst
@@ -38,9 +38,9 @@ General Notes
 
      apk add --virtual build-deps file make gcc musl-dev libffi-dev
 
-  On Fedora Rawhide for Fedora 33, one might need to do this::
+  On Fedora Rawhide or Fedora 33, one might need to do this::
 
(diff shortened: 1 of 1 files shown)


jamadden:
Do you know what version that changed it? It would be nice to note "use yum on XXX and after XXX use dnf".

fabaff:
`dnf` is the default since Fedora 22 (released in 2015) and also for CentOS 8/RHEL 8.

docs/development/installing_from_source.rst
```diff
@@ -38,9 +38,9 @@ General Notes
 
      apk add --virtual build-deps file make gcc musl-dev libffi-dev
 
-  On Fedora Rawhide for Fedora 33, one might need to do this::
+  On Fedora Rawhide or Fedora 33, one might need to do this::
```
jamadden, line 41:
```suggestion
  On Fedora Rawhide for Fedora 33, one might need to do this::
```

I believe the "for" was correct there. This is documenting one specific system version, not all versions of Rawhide.

PR #1690, opened 2020-10-21
Update package name

fabaff:
The `python3` package installs Python 3.9. Also, Fedora Rawhide will become F34.

### DIFF
--- a/scripts/releases/make-manylinux
+++ b/scripts/releases/make-manylinux
@@ -50,8 +50,8 @@ if [ -d /gevent -a -d /opt/python ]; then
     ls -ld /cache
     ls -ld /cache/pip
     yum -y install libffi-devel ccache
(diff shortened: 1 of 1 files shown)


jamadden:
Thanks for the suggestion. That's just an internal comment in an internal script, though, serving as one example. I expect people needing to read it will be able to adapt it as appropriate to their system.

fabaff:
> I expect people needing to read it will be able to adapt it as appropriate to their system.

Still there is no `yum` and not package called `python39`.

jamadden:
That's fine, it doesn't matter. This is not a Fedora script meant to run on any current Fedora distribution (it runs only in the specific manylinux environment), those are not executed instructions, they were a comment documenting the steps needed to test this *outside* that environment in *one specific other environment*. To test them elsewhere, adapt them.

PR #1749, opened 2021-01-18
Add note about dropping building some binary wheels

mik-laj:
Hello,

This change is not obvious to everyone, so I add a note in the Changelog about it.
Related: https://github.com/gevent/gevent/issues/1748

CC: @jamadden 

Best regards,
Kamil Breguła

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,11 @@ installed. Reported by Josh Zuech.
 21.1.0 (2021-01-15)
 ===================
 
+Changes
+-------
+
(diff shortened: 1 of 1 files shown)


jamadden:
As 3.6 is now in security fix only mode, it is not an important platform for new gevent releases; expect all gevent support for it to end soon. 

Note that binary wheels haven’t been provided for the last three releases; this is not a new absence. 

Thank you for the suggested change, but given the above I don’t think I’ll be accepting it. (Binary wheels are nothing but a convenience, not a guarantee. )

PR #1750, opened 2021-01-19
Add docs for use gevent.monkey as module

aisk:


jamadden:
Thanks,  but this is already documented. See the end of the docstring.

aisk:
Hi @jamadden , this is already documented, but not so eye-catching, I'm using gevent for years and been hired by some companies and some of them has thousands more developers, but none of them know this option, and always struggle by the monkey patch occasion.

So I think we should improve the document and make this more eye-catching?

jamadden:
There's an entire section, with a bold header, and its own URL (http://www.gevent.org/api/gevent.monkey.html#use-as-a-module) already.  That's much more eye-catching than another single paragraph.

Untitled

aisk:
@jamadden sorry, I missed it, I think it's ok, but still think it will be more good if we recommended using the module way to monkey patch user's program, or at least write it besides the coding way to monkey patch. You know people (like me) always stop read the docuement if they think they found the way to solve the problem.

jamadden:
That's *an* option, but certainly not the *recommended* option. Coding it directly is always preferred when possible.

PR #1789, opened 2021-04-29
Fix build for IBM i (remove libperfstat linking)

ThePrez:
IBM i is an AIX derivative and even identifies 'AIX` in `sys.platform`. 
However, IBM i does not have or use a `perfstat` library, so builds got broken here: https://github.com/gevent/gevent/commit/b9ed4e2130d9b15a9d040f1346a76aeb80c038fa


jamadden:
Thanks! I have no way to test or verify this, but I see you've [contributed in a similar area to greenlet](https://github.com/python-greenlet/greenlet/pull/197) so I'm happy to take your word for it.

PR #1814, opened 2021-08-25
Link the libuv `_corecffi.pyd` to the minor-version Python DLL on Windows.

ClaudioHoffmann:
All other gevent `.pyd` files do the same, and it avoids crashes in virtual environments that don't provide support for the limited Python API via `python3.dll`.

This would fix such a crash for us, so a new PyPI release with this fix would be much appreciated.

### DIFF
--- a/src/gevent/libuv/_corecffi_build.py
+++ b/src/gevent/libuv/_corecffi_build.py
@@ -275,6 +275,14 @@ elif sys.platform.startswith('aix'): # pragma: no cover
     if os.uname().sysname != 'OS400':
         _add_library('perfstat')
 elif WIN:
+    # All other gevent .pyd files link to the specific minor-version Python
+    # DLL, so we should do the same here. In virtual environments that don't
+    # contain the major-version python?.dll stub, _corecffi.pyd would otherwise
+    # cause the Windows DLL loader to search the entire PATH for a DLL with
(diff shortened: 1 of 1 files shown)


jamadden:
I'm not sure about this for a couple reasons. Using the limited API/ABI is generally considered a positive state of affairs, one that gevent should be getting closer to, not farther from (because of the reduction in required maintenance).  CFFI already uses the limited API/ABI where possible, and this flag to disable that is not documented and seems to be intended for internal use only.

Why would a modern virtual environment (post 3.5) not include the limited ABI DLL?

ClaudioHoffmann:
> Why would a modern virtual environment (post 3.5) not include the limited ABI DLL?

[Oversight, mainly.](https://github.com/Slicer/Slicer/issues/5816) I agree that it's a bug in that environment, but it wouldn't have caused a crash if all of gevent consistently linked to either `python3.dll` or `python3*.dll`. So in the ideal case, the issue should be fixed on both ends.

> I'm not sure about this for a couple reasons. Using the limited API/ABI is generally considered a positive state of affairs, one that gevent should be getting closer to, not farther from (because of the reduction in required maintenance).

I was under the impression that this would be the least invasive fix, since [you're already providing distinct packages for all supported Python versions](https://pypi.org/project/gevent/#files). That makes it seem like changing everything else to the limited API would be a much larger task.

PR #1835, opened 2021-11-20
Fix typos

kianmeng:


jamadden:
Thank you for this PR. I cannot accept it as is because it changes files in src/greentest. With only one exception, we don't change those files. They are direct copies from upstream.

kianmeng:
@jamadden Sorry, caught up with other PRs. Okay to resend PR again?

Issue #1412, opened 2019-04-24
Dev installs broken with pip 19.1


jamadden:
* gevent version: master

With the newly released pip 19.1, the existence of a pyproject.toml breaks the use of editable installs, making it sadly essentially useless to developers (I *just* added it😿 ). As observed [on Appveyor](https://ci.appveyor.com/project/denik/gevent/builds/24078529/job/8d8anh69srmdqq0e#L155):

```
if not "%GWHEEL_ONLY%"=="true" %PYEXE% -m pip install -U --upgrade-strategy=eager -r dev-requirements.txt

ERROR: Error installing 'file:///C:/projects/gevent (from -r dev-requirements.txt (line 18))': 
editable mode is not supported for pyproject.toml-style projects. This project is being 
processed as pyproject.toml-style because it has a pyproject.toml file with a 
"build-backend" key in the "build_system" value. See PEP 517 for the relevant specification.
```

See [this issue](https://github.com/pypa/pip/issues/6314#issuecomment-469176276):

> I'd take a hardline view here. Given that PEP 517 doesn't support editable installs, projects shouldn't be using PEP 517 if they want to support them. If this is a problem, then the solution should be for someone to do the work of defining a PEP 517 compatible standard for editable installs. Then pip (and any other frontends that are developed) can support it cleanly for any backend that provides editable install support.

jamadden:
cf https://github.com/pypa/pip/issues/6434, discussing steps forward

jamadden:
Supposedly the next bugfix release of pip (19.1.1) will make `--no-use-pep517` do something compatible here.

PR #1836, opened 2021-11-25
docs(dev): Removes no-longer useful warning about pep 19.1

doublethefish:
The underlying issue is fixed in pip v21 (at least) and it is reasonable
to expect people to be on some good-working pip version now.

The warning is also a false-negative/misleading when you're looking at
install and deps issues as it wastes time, suggesting pip is the issue.

### DIFF
--- a/docs/development/getting_started.rst
+++ b/docs/development/getting_started.rst
@@ -51,11 +51,6 @@ can be done in one line by installing the ``dev-requirements.txt`` file::
 
   (gevent-env) $ pip install -r dev-requirements.txt
 
-.. warning::
-
-   This pip command does not work with pip 19.1. Either use pip 19.0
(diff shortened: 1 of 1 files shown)


doublethefish:
Also see #1412.

jamadden:
Thanks for the suggestion, but I'm not sure it is reasonable to assume people always have an up-to-date pip. Previous experience when trying to use newer packaging features here suggests there's a significant long tail of environments that take multiple years to update. 

We can look at this again when we drop Python 2.7 and 3.6 support, probably sometime in the first or second quarter of next year.

doublethefish:
The comment, as it is, is misleading.

jamadden:
It seems pretty specific to me, but I'm open to an alternate wording?

PR #1844, opened 2021-12-07
Update to testing 3.10.0 final

jamadden:
That's the latest that GHA currently has. We've been testing 3.10.0 on AppVeyor and in the manylinux images.


auvipy:
can we expect a new release before holiday session?

PR #1845, opened 2021-12-11
Update tested versions of PyPy

### LINKED ISSUE #1843: Update PyPy support
On GHA, we're currently testing PyPy's 3.6 release, which has been supplanted by a 3.7 and (still beta, I think) 3.8 releases.

On AppVeyor, we're testing a 32-bit version of 2.7 (7.3.1); there are newer versions of 2.7, but it's possible that they are only 64-bit (which seems fine to me). (On GHA, we're testing 7.3.6/64-bit.)

jamadden:
Fixes #1843.


mattip:
Thanks!

PR #1859, opened 2022-01-18
Improved documentation of link() and rawlink()

aathan:
I think this is probably a common question because link() and rawlink() must often be called after a greenlet is started.


aathan:
This pull request is for documentation changes only. Not sure why the testing failed ... appears to be unrelated to the change.

jamadden:
The tests hit an intermittent kernel bug on macOS. 

Thanks for the PR.

PR #1863, opened 2022-01-28
Force "fork" multiprocessing context for test__core_fork

mgorny:
Use the "fork" multiprocessing context explicitly within test__core_fork
as the test does not work correctly when using the "spawn" context.
The latter is likely to become the default in PyPy3.9 due to the issues
that the "fork" method causes with threading.


arcivanov:
Fork is not available on Windows.

arcivanov:
Additionally MP doesn't work with Gevent at all as is, unless used with a project such as https://github.com/karellen/geventmp/

PR #1864, opened 2022-02-02
manylinux: Truly disable the effect of -Ofast/-ffast-math

ento:
## Context
Commit history and issue history leading up to this PR:

- https://github.com/gevent/gevent/commit/59478ebb12384403932125d9ffefea4413840ca6 added `-Ofast` to `CFLAGS`
- #1820 added `-fno-fast-math` to `CFLAGS` in order to prevent process-wide floating-point flags from being set as an effect of `-Ofast`, which enables `-ffast-math`
- It was discovered in https://github.com/HypothesisWorks/hypothesis/issues/3092 that importing gevent still enables the floating-point flags

## Minimally reproducible example

Running `mxcsr_test/test.sh` from my [sscce-no-fast-math branch](https://github.com/ento/gevent/compare/master...sscce-no-fast-math) prints the status of floating-point flags after importing gevent (version 21.12.0):

```
$ ./test.sh
Requirement already satisfied: gevent==21.12.0 in ./env/lib/python3.10/site-packages (from -r requirements.txt (line 1)) (21.12.0)
Requirement already satisfied: zope.interface in ./env/lib/python3.10/site-packages (from gevent==21.12.0->-r requirements.txt (line 1)) (5.4.0)
Requirement already satisfied: setuptools in ./env/lib/python3.10/site-packages (from gevent==21.12.0->-r requirements.txt (line 1)) (56.0.0)
Requirement already satisfied: greenlet<2.0,>=1.1.0 in ./env/lib/python3.10/site-packages (from gevent==21.12.0->-r requirements.txt (line 1)) (1.1.2)
Requirement already satisfied: zope.event in ./env/lib/python3.10/site-packages (from gevent==21.12.0->-r requirements.txt (line 1)) (4.5.0)
running build_ext
copying build/lib.linux-x86_64-3.10/mxcsr.cpython-310-x86_64-linux-gnu.so -> 
daz_enabled=True
ftz_enabled=True
Did not expect DAZ nor FTZ to be enabled
```

## Why does not `-fno-fast-math` take effect?

gcc's [crtfastmath.c](https://github.com/gcc-mirror/gcc/blob/releases/gcc-11.2.0/libgcc/config/i386/crtfastmath.c#L95-L96) is the code responsible for enabling the floating-point flags, and it appears to be that it gets linked whenever `-Ofast`, `-ffast-math`, or `-funsafe-math-optimizations` is specified: https://github.com/gcc-mirror/gcc/blob/releases/gcc-11.2.0/gcc/config/i386/gnu-user-common.h#L50

Running `cd gcc_test; make; cat result.txt` in my [sscce-no-fast-math branch](https://github.com/ento/gevent/compare/master...sscce-no-fast-math) demonstrates that `-fno-fast-math` cancels out `-ffast-math` but not `-Ofast` when it comes to preventing `crtfastmath.c` from being linked.

```
$ cat result.txt 
./check_ofast
DAZ enabled: yes
FTZ enabled: yes
./check_fm
DAZ enabled: yes
FTZ enabled: yes
./check_ofast_nofm
DAZ enabled: yes
FTZ enabled: yes
./check_fm_nofm
DAZ enabled: no
FTZ enabled: no
./check_ofast_fm_nofm
DAZ enabled: yes
FTZ enabled: yes
```

`./check_ofast` is an executable compiled with `-Ofast`, `./check_fm` with `-ffast-math`, `./check_ofast_nofm` with `-Ofast -fno-fast-math`, and so on.

## This PR's changes

From gcc's documentation: https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html

> -Ofast [..] enables optimizations that are not valid for all standard-compliant programs
 
The commit that introduced `-Ofast`, 59478ebb12384403932125d9ffefea4413840ca6, only mentions build time, and enabling this level of optimization that can break standard-compliance didn't seem to be the intention. So this PR sets the optimization level to `-O3` to avoid any other unexpected behavior.

The behavior of `-fno-fast-math` described above may be a bug on gcc's end, but I figured we can simply go with `-O3` here.

### DIFF
--- a/scripts/releases/make-manylinux
+++ b/scripts/releases/make-manylinux
@@ -36,8 +36,8 @@ export CCACHE_DIR="/ccache"
 GEVENT_WARNFLAGS="-Wno-strict-aliasing -Wno-comment -Wno-unused-value -Wno-unused-but-set-variable -Wno-sign-compare -Wno-parentheses -Wno-unused-function -Wno-tautological-compare -Wno-strict-prototypes -Wno-return-type -Wno-misleading-indentation"
 OPTIMIZATION_FLAGS="-pipe"
 if [ "$DOCKER_IMAGE" == "quay.io/pypa/manylinux2014_aarch64" -a -n "$GITHUB_ACTIONS" ]; then
-    # Compiling with -Ofast on the arm emulator takes hours. The default settings have -O3,
-    # and adding -Os doesn't help much. So maybe -O1 will.
+    # Compiling with -Ofast is a no-go because of the regression it causes (#1864).
+    # The default settings have -O3, and adding -Os doesn't help much. So maybe -O1 will.
     echo "Compiling with -O1"
     OPTIMIZATION_FLAGS="-pipe -O1"
     SLOW_ARM=1
--- a/scripts/releases/make-manylinux
+++ b/scripts/releases/make-manylinux
@@ -43,14 +43,6 @@ if [ "$DOCKER_IMAGE" == "quay.io/pypa/manylinux2014_aarch64" -a -n "$GITHUB_ACTI
     SLOW_ARM=1
 else
     echo "Compiling with -O3"
-    # Caution against using -Ofast:
-    # -Ofast includes -ffast-math which affects process-wide floating-point flags (e.g. can affect numpy).
-    # -fno-fast-math disables *some* features of -ffast-math, but as long as -Ofast is specified,
-    # it still affects floating-point flags:
-    # https://github.com/gcc-mirror/gcc/blob/releases/gcc-11.2.0/gcc/config/i386/gnu-user-common.h#L50
-    #
-    # We *could* turn on other flags that are included in -Ofast individually, but for gevent's usecase, it's not
-    # a priority to enable aggressive optimizations at the expense of possibly violating language standards.
     OPTIMIZATION_FLAGS="-pipe -O3"
 fi
 export CFLAGS="$OPTIMIZATION_FLAGS $GEVENT_WARNFLAGS"
--- a/scripts/releases/make-manylinux
+++ b/scripts/releases/make-manylinux
@@ -42,10 +42,16 @@ if [ "$DOCKER_IMAGE" == "quay.io/pypa/manylinux2014_aarch64" -a -n "$GITHUB_ACTI
     export CFLAGS="-O1 $GEVENT_WARNFLAGS"
     SLOW_ARM=1
 else
-    echo "Compiling with -Ofast"
-    # Note: -Ofast includes -ffast-math which affects process-wide floating-point flags (e.g. can affect numpy).
-    #       We opt out of -ffast-math explicitly. Other libraries can still trigger it.
-    export CFLAGS="-Ofast -fno-fast-math $GEVENT_WARNFLAGS"
+    echo "Compiling with -O3"
+    # Caution against using -Ofast:
+    # -Ofast includes -ffast-math which affects process-wide floating-point flags (e.g. can affect numpy).
+    # -fno-fast-math disables *some* features of -ffast-math, but as long as -Ofast is specified,
+    # it still affects floating-point flags:
+    # https://github.com/gcc-mirror/gcc/blob/releases/gcc-11.2.0/gcc/config/i386/gnu-user-common.h#L50
+    #
+    # We *could* turn on other flags that are included in -Ofast individually, but for gevent's usecase, it's not
+    # a priority to enable aggressive optimizations at the expense of possibly violating language standards.
+    export CFLAGS="-O3 $GEVENT_WARNFLAGS"
 fi
 # -lrt: Needed for clock_gettime libc support on this version.
 # -pthread: Needed for pthread_atfork (cffi).
--- a/scripts/releases/make-manylinux
+++ b/scripts/releases/make-manylinux
@@ -33,13 +33,13 @@ export BUILD_LIBS=$HOME/.libs
 export CCACHE_DIR="/ccache"
 # Disable some warnings produced by libev especially and also some Cython generated code.
 # Note that changing the value of these variables invalidates configure caches
-GEVENT_WARNFLAGS="-pipe -Wno-strict-aliasing -Wno-comment -Wno-unused-value -Wno-unused-but-set-variable -Wno-sign-compare -Wno-parentheses -Wno-unused-function -Wno-tautological-compare -Wno-strict-prototypes -Wno-return-type -Wno-misleading-indentation"
-export CFLAGS="$GEVENT_WARNFLAGS"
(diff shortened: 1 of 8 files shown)


amitpl:
@ento we are also hitting the same assertion error while using gevent+hypothesis+locust. Do you know of any workaround until this PR is released?

ento:
@amitpl In my case, I'm 'working around' it by holding off on merging the feature branch that introduces gevent (as a transitive dependency of locust, as it turns out)  to a codebase that uses hypothesis. 

Does unsetting `daz` and `ftz` work in a non-Docker environment? I'd try [adding this little C extension that reads the current state of the flags](https://github.com/ento/gevent/compare/master...sscce-no-fast-math#diff-84836f145e1d0f532971f7dedb0f731e8e1a7ddb8f041043ba88d34de7a56e1a) to the codebase and verify that they really are unset right before the assertion error gets thrown, patching hypothesis' internal code to call `mxcsr.daz_enabled()` and `mxcsr.ftz_enabled()` and print the results if needed.

buchi-busireddy:
Thanks for the PR @ento 
FTR, we are also running into this issue and even trying `daz` and `ftz` didn't work inside Docker container so we are fully blocked on using the Locust library (since it uses `gevent` internally).

Just curious, any rough timelines on when this could be merged and released?

ml31415:
@jamadden Would you mind to have a look at this? Would be great if we could get it merged :)

tgerdesnv:
Any update on this?

ikonst:
Aww!
```console
$ echo 'int main() {}' > foo.c
$ gcc -Ofast -fno-fast-math -o foo-gcc foo.c
$ clang -Ofast -fno-fast-math -o foo-clang foo.c
```
and then
```console
$ objdump -d foo-clang | grep fast
00000000004003e0 <set_fast_math>:
$ objdump -d foo-gcc | grep fast
00000000004003f0 <set_fast_math>:
```
clang is apparently [intentionally matching](https://github.com/llvm/llvm-project/blob/main/clang/lib/Driver/ToolChain.cpp#L1038) gcc behavior here.

@jamadden mind if we part with -Ofast?

scripts/releases/make-manylinux
```diff
@@ -33,20 +33,27 @@ export BUILD_LIBS=$HOME/.libs
...
-    export CFLAGS="-Ofast -fno-fast-math $GEVENT_WARNFLAGS"
+    echo "Compiling with -O3"
+    # Caution against using -Ofast:
+    # -Ofast includes -ffast-math which affects process-wide floating-point flags (e.g. can affect numpy).
+    # -fno-fast-math disables *some* features of -ffast-math, but as long as -Ofast is specified,
+    # it still affects floating-point flags:
+    # https://github.com/gcc-mirror/gcc/blob/releases/gcc-11.2.0/gcc/config/i386/gnu-user-common.h#L50
+    #
+    # We *could* turn on other flags that are included in -Ofast individually, but for gevent's usecase, it's not
+    # a priority to enable aggressive optimizations at the expense of possibly violating language standards.
```
ikonst, line 53: I'd probably remove this comment. We don't need to mention what _not to do_ at every place where we _do_ something. A regression test would be a better way to codify this.

ml31415:
Congratulations, this bug made public news appearance now, since there is no one there to press the button to get this accepted :( Any chance to get gevent into PSF hands for example, to guarantee some more maintainer attention for such occasions?

https://www.golem.de/news/fpu-tausende-python-pakete-koennten-falsche-berechnungen-liefern-2209-168134.html

ikonst:
Not exactly world news :)

Jason has been maintaining this for about a decade. Just because this concern is not being addressed doesn't mean that it's abandoned.

Also, does the PSF provide guaranteed maintainers for projects? Take a look at the code, gevent is pretty complex.

scripts/releases/make-manylinux
ento: I like the idea of a regression test, and thanks for the actual test code!

Why is it alright to skip the test under that condition? In my environment, the skip condition is met, but the assertion fails.

```python
$ python
Python 3.7.12 (default, Sep  4 2021, 05:47:15) 
[GCC 10.3.0] on linux
>>> import gevent
>>> import sys
>>> sys.float_info.max == 1.7976931348623157e+308
True
>>> assert 5e-324 != 0
AssertionError
```

ml31415:
I'm very grateful for Jasons work and the code is surely complex. Nevertheless, it shouldn't happen that a project of that size depends on a single maintainer. At least some co-maintainer to accept occasional bug fixing PRs should be there. It's not that gevent would need new features at this point.

scripts/releases/make-manylinux
ikonst: Err, sorry, flip the skip condition :) `==` -> `!=`

The test is whether a subnormal floating point value is equal to zero. What is subnormal depends on the size of float on the platform -- that's why I'm comparing to `float_info.max`: meant for this to be a test for 64-bit platforms.

ikonst: Actually, starting Python 3.9 there's a more portable way to get it, that should obliterate the need for any `skipif`s:
```python
import math

assert math.nextafter(0, 1) != 0
```

ento: Ah, thanks. Tried adding the test in a separate branch, which got messy from trying to fix CI failures coming from newer versions of pylint and greenlet. I'll create a separate PR for those fixes and push the test to this branch.

The test did fail in CI (revealing that I missed updating `CFLAGS` used in CI) https://github.com/ento/gevent/runs/8286207244?check_suite_focus=true

ento: A regression test has been pushed to this branch, and https://github.com/gevent/gevent/pull/1906 attempts to fix CI failures.

src/gevent/tests/test__issue1864.py
```diff
@@ -0,0 +1,21 @@
...
+    """
+    Enabling the -Ofast compiler flag resulted in subnormal floats getting
+    disabled the moment when gevent was imported. This impacted libraries
+    that expect subnormal floats to be enabled.
+    """
+    @unittest.skipUnless(PY39, "Need math.nextafter()")
+    def test_subnormal_is_not_zero(self):
+        import gevent
+
+        assert math.nextafter(0, 1) != 0
```
Zac-HD, line 17: There's a less elegant way to test this on Python 3.8 and earlier too - [`sys.float_info.min` is the minimum *normalized* float](https://docs.python.org/3/library/sys.html#sys.float_info), so you can `assert (sys.float_info.min / 2) > 0` and with a block comment it's not even too confusing.

ikonst: @Zac-HD, that's great. Makes perfect sense, and somewhat clearer than my suggestion.

ento: Thanks for the suggestion. Updated the test to use `sys.float_info.min` (no `skipUnless` anymore!), and confirmed that [it fails as expected when gevent is built with `-Ofast`](https://github.com/ento/gevent/actions/runs/3123769165).

jamadden:
Thank you all for your patience and contributions! This has been merged as-of https://github.com/gevent/gevent/commit/e29bd2ee11ca5f78cc9c9bc784d875681afb8d23

Issue #1685, opened 2020-10-06
RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility


thiras:
* gevent version: 1.4.0
* gunicorn version: 20.0.4
* Python version: cPython 3.7.3 (3.7.3-1 from Debian Buster repository)
* Operating System: Debian Buster 10 4.19.0-11-amd64

### Description:

I was trying to deploy [ruqqus](https://github.com/ruqqus/ruqqus/).

```python
/usr/lib/python3.7/importlib/_bootstrap.py:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
  return f(*args, **kwds)
/usr/lib/python3.7/importlib/_bootstrap.py:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
  return f(*args, **kwds)
/usr/lib/python3.7/importlib/_bootstrap.py:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
  return f(*args, **kwds)
/usr/lib/python3.7/importlib/_bootstrap.py:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
  return f(*args, **kwds)
/usr/lib/python3.7/importlib/_bootstrap.py:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
  return f(*args, **kwds)
/usr/lib/python3.7/importlib/_bootstrap.py:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
  return f(*args, **kwds)
/usr/lib/python3.7/importlib/_bootstrap.py:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
  return f(*args, **kwds)
/usr/lib/python3.7/importlib/_bootstrap.py:219: RuntimeWarning: greenlet.greenlet size changed, may indicate binary incompatibility. Expected 144 from C header, got 152 from PyObject
  return f(*args, **kwds)
Segmentation fault
```

### What I've run:

[...]

PR #1869, opened 2022-02-10
WIP: Fix #1865 by using importlib instead of pkg_resources

### LINKED ISSUE #1865: Popen stall in ThreadPoolEecutor under docker
* gevent version:
21.12.0, wheels from pypi
* Python version: python 3.10.2
* Operating System: docker image [3.10.2-slim-bullseye](https://github.com/docker-library/python/blob/e0dacc4c580f4ad0aa68136789add4e37bdaab91/3.10/slim-bullseye/Dockerfile)

### Description:

Hi, the problem appears suddenly today. Build stage has stall on `manage.py compilemessages`. `strace` shows only a infinite futex waiting.
Usually it's always worked ok. I don't know why exactly a behaviour is changed. But I found why it is stall.

Simple reproducible example ((not) worked only under docker)

```python
from gevent import monkey
import gevent

if __name__ == '__main__':
    monkey.patch_thread();
    monkey.patch_subprocess();

    import subprocess
    from concurrent.futures import ThreadPoolExecutor

    def func():
        # gevent.sleep(0) # uncomment to fail
        return subprocess.run(["ls", "-l", "/dev/null"], capture_output=True)

    with ThreadPoolExecutor(max_workers=5) as executor:
        fut = executor.submit(func)
        print(fut.result())
```

This code will stall, but ctr+c give traceback:

```python-traceback
^CException ignored in: <built-in method acquire of _thread.lock object at 0x7f1af2ab8800>
  File "/usr/local/lib/python3.10/site-packages/gevent/os.py", line 431, in fork_and_watch
    pid = fork()
KeyboardInterrupt:
  File "/var/www/cabinet/test.py", line 16, in <module>
    fut = executor.submit(func)
  File "/usr/local/lib/python3.10/concurrent/futures/thread.py", line 162, in submit
    with self._shutdown_lock, _global_shutdown_lock:
RuntimeError: release unlocked lock
```

If uncomment sleep in `func` all works fine. So problem occurs due to the code is not concurrent really. 
`ThreadPoolExecutor` acquires a several locks in [submit](https://github.com/python/cpython/blob/3.10/Lib/concurrent/futures/thread.py#L162). CPython calls a some handlers before `fork`  [here](https://github.com/python/cpython/blob/3.10/Modules/posixmodule.c#L574). Particularly `concurrent.futures.thread` sets [one](https://github.com/python/cpython/blob/3.10/Lib/concurrent/futures/thread.py#L41). As the code are not runned concurrently and `_global_shutdown_lock` is the simple Lock, not RLock, so it will be deadlocked.

maciejp-ro:
Freezes in #1685 are caused by new versions of `setuptools` (which is imported by `gevent` early on) vendoring `more_itertools`, which imports `concurrent.futures` before it can be monkey-patched.

Setuptools 60.8.2 fixed that import, but `pkg_resources` is deprecated and should be replaced by `importlib` or appropriate backport.

This is the replacement suggested by setuptools maintainers at https://github.com/pypa/setuptools/issues/3090#issuecomment-1033167013 – it uses stdlib on Python 3.10 and a flexible backport in earlier versions. This is supposed to work in all Python versions supported by gevent.

So far, I was only able to test it on Python 3.9 & 3.10 (and confirm that it fixes the freeze with faulty versions of setuptools). It times out in doctests, which doesn't seem related. I'll try to run more tests when I switch to my other computer.

`pkg_resources` is still used in the test suite. I can work on replacing it there too, but I'd love to hear some feedback first.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -226,10 +226,9 @@ install_requires = greenlet_requires + CFFI_REQUIRES + [
     # ultimately be published, but at this writing only the event
     # interfaces are.
     'zope.interface',
-    # setuptools is also used (via pkg_resources) for event
-    # notifications. It's a hard dependency of zope.interface
-    # anyway.
-    'setuptools',
+    # Backport of importlib.metadata.entry_points for older version of Python,
+    # lets us avoid deprecated pkg_resources
+    "backports.entry-points-selectable; python_version<'3.10'"
 ]
 
 # We use headers from greenlet, so it needs to be installed before we
@@ -434,6 +433,9 @@ def run_setup(ext_modules):
                 # The backport for contextvars to test patching. It sadly uses the same
                 # import name as the stdlib module.
                 'contextvars == 2.4 ; python_version > "3.0" and python_version < "3.7"',
+
+                # For pkg_resources until we fully switch to importlib
+                'setuptools',
             ],
         },
         # It's always safe to pass the CFFI keyword, even if
--- a/src/gevent/events.py
+++ b/src/gevent/events.py
@@ -63,6 +63,7 @@ __all__ = [
 
 # pylint:disable=no-self-argument,inherit-non-class
 import platform
+import sys
 
 from zope.interface import Interface
 from zope.interface import Attribute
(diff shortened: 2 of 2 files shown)


jamadden:
Thanks for this PR. This work was done as part of the larger work of supporting Python 3.12.

PR #1877, opened 2022-03-26
fix some typos

cuishuang:


jamadden:
Thank you for the PR. However, we cannot accept any changes in the `deps/` or `src/greentest/` directory as those are vendored from upstream. Please feel free to open a PR containing only changes to gevent's code.

PR #1879, opened 2022-04-06
Improve format run info

aathan:
This adds `include_ready` parameter to format_run_info() etc so that the dump can exclude greenlets that are ready. Also, changed the output from labeling such greenlets as `; finished` to labeling them `; ready (finished)` since the code's terminology for a finished greenlet is that it is "ready". Also, the dummy function `_ready()` is renamed `_return_false()` to better self-document its purpose.

### DIFF
--- a/src/gevent/util.py
+++ b/src/gevent/util.py
@@ -33,7 +33,7 @@ _STACK_LIMIT = 20 if PYPY else None
 def _noop():
     return None
 
-def _ready():
+def _return_false():
     return False
 
 class wrap_errors(object):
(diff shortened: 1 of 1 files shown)


aathan:
Oh, I'll note I took a brute-force approach to this change. I didn't have time to fully grok the tree stuff, and a better implementation might avoid gathering all the dead/finished greenlets in the firstplace.

aathan:
The failed jobs seem, at first blush, unrelated to the changes in the PR.

PR #1896, opened 2022-07-01
Fix Intel compiler deprecated options

stephenmsachs:
Options `-we ###` and `-wd ###` should not include a whitespace. They are also deprecated and `-diag-error` and `-diag-disable` are their replacements. 
Intel compiler 2021.6 is not able to be used in configure without the proposed patch.

### DIFF
--- a/deps/c-ares/configure
+++ b/deps/c-ares/configure
@@ -21736,8 +21736,8 @@ squeeze() {
       INTEL_UNIX_C)
         #
                         tmp_CFLAGS="$tmp_CFLAGS -std=gnu89"
-                                                tmp_CPPFLAGS="$tmp_CPPFLAGS -we 140,147,165,266"
-                                        tmp_CPPFLAGS="$tmp_CPPFLAGS -wd 279,981,1469"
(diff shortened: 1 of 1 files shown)


jamadden:
Thank you for your interest and this PR! Unfortunately, it's not something I'm going to be able to accept.

c-ares is a third-party library, and gevent just includes a copy of the latest release for convenience. That means we don't make any changes to the code unless absolutely necessary, because that imposes a heavy maintenance burden as we recieve updates from upstream. 

In extreme cases we can, and have, done so, but we need to do so using patch files; see the README in the `deps` directory.

Because c-ares is an optional extension, and because you can link to a c-ares you've built yourself instead of using the one included in the gevent source, I wouldn't consider this an "extreme" circumstance.

I would encourage you to take this problem upstream to the [c-ares project](https://c-ares.org). Once it is fixed upstream and released, we can incorporate the change here. (Note that `configure` is auto-generated from `configure.ac` and a bunch of M4 macro files from GNU autotools; it's possible that the problem is even further upstream than c-ares. But I don't know enough about autotools or how they use it to be sure.)

PR #1905, opened 2022-09-07
Stop building with -Ofast

ikonst:
Stop building with `-Ofast` to prevent altering the global state of the process with regard to floating point precision (see https://simonbyrne.github.io/notes/fastmath/).

Previous attempt in https://github.com/gevent/gevent/pull/1820 to keep `-Ofast` while adding `-fno-fast-math` proved futile, as both on gcc and clang ([ref](https://github.com/llvm/llvm-project/blob/fb9fc79809d5c2c3ab8809f7ff98fbfa8c3a9e7e/clang/lib/Driver/ToolChain.cpp#L1038-L1039)) `-Ofast` would enable "fast math" regardless.

```console
$ echo 'int main() {}' > foo.c
$ gcc -Ofast -fno-fast-math -o foo-gcc foo.c
$ clang -Ofast -fno-fast-math -o foo-clang foo.c
```
and then
```console
$ objdump -d foo-clang | grep fast
00000000004003e0 <set_fast_math>:
$ objdump -d foo-gcc | grep fast
00000000004003f0 <set_fast_math>:
```

Closes #1864.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,7 @@ env:
   # These are shared between GCC and clang so it must be a minimal set.
   # TODO: Figure out how to set env vars per platform without resorting to inline scripting.
   # Note that changing the value of these variables invalidates configure caches
-  CFLAGS: -Ofast -pipe -Wno-strict-aliasing -Wno-comment
+  CFLAGS: -O2 -pipe -Wno-strict-aliasing -Wno-comment
   CPPFLAGS: -DEV_VERIFY=1
   # Uploading built wheels for releases.
   # TWINE_PASSWORD is encrypted and stored directly in the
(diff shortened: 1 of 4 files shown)


ikonst:
p.s. came up again here: https://moyix.blogspot.com/2022/09/someones-been-messing-with-my-subnormals.html

FWIW, gevent isn't the only offender in the Python ecosystem. The root cause is the compiler feature.

ml31415:
Have a look here, this is already waiting for a long time to be accepted. Maintainers seem to have left the project :(((

https://github.com/gevent/gevent/pull/1864

ikonst:
@ml31415 thanks, it somehow eclipsed that was a PR and not an issue :)

PR #1906, opened 2022-09-11
Fix various build and test failures

ento:
While working on #1864, I noticed that CI probably won't pass even on the master branch. This PR will hopefully fix all the issues.

* An alpha version of greenlet's next major version has been released. The step that installs greenlet in CI didn't specify an upper limit, and since the alpha version isn't backwards compatible, [builds were failing](https://github.com/ento/gevent/runs/8282988158?check_suite_focus=true)
* Newer versions of pylint has been released, which include things like new checks and changes to extensions that are enabled by default
* PyPy's class hierarchy around SSL-related errors may have changed
* setuptools' new implementation of editable installs broke Appveyor builds it seems

[Appveyor builds were failing](https://ci.appveyor.com/project/ento/gevent/builds/44745934/job/70myi721h3dd98px) from

>  Creating library C:\Users\appveyor\AppData\Local\Temp\1\tmpucf7nr76.build-temp\Release\deps/c-ares/src/lib\cares.cp310-win_amd64.lib and object C:\Users\appveyor\AppData\Local\Temp\1\tmpucf7nr76.build-temp\Release\deps/c-ares/src/lib\cares.cp310-win_amd64.exp
>  ares__addrinfo_localhost.obj : error LNK2001: unresolved external symbol __imp_FreeMibTable
>  ares__addrinfo_localhost.obj : error LNK2001: unresolved external symbol __imp_GetUnicastIpAddressTable
>  cares.obj : error LNK2001: unresolved external symbol hstrerror

Previously released wheels for Windows don't include any DLL for `cares`, so it seems like it was trying to build `c-ares` when it didn't need to. The first two errors went away when I added `iphlpapi` to `LIBRARIES` in `_setuputils.py`, but it seems like `hstrerror` ([invoked here](https://github.com/gevent/gevent/blob/5858d226045f701deb8e4b49c03db4cab63053d6/src/gevent/resolver/cares.pyx#L211)) is something unix-only and a [Windows-specific API needs to be used](https://stackoverflow.com/a/57137600/20226).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -258,7 +258,7 @@ jobs:
           pip install -q -U 'faulthandler; python_version == "2.7" and platform_python_implementation == "CPython"'
           pip install -q -U 'cffi;platform_python_implementation=="CPython"'
           pip install -q -U 'cython>=3.0a9'
-          pip install 'greenlet>=1.0a1;platform_python_implementation=="CPython"'
+          pip install 'greenlet>=1.0a1,<2;platform_python_implementation=="CPython"'
 
       - name: Build gevent
         run: |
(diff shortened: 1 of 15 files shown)


jamadden:
Thank you!

PR #1914, opened 2022-10-14
Switch to the non-legacy setuptools backend

### LINKED ISSUE #1910: pyproject.toml uses non-public __legacy__  setuptools backend
* gevent version: git (as of beb98ce4742181caaa6cec9cedecaaa2a04f910b)
* Python version: n/a
* Operating System: Gentoo Linux

### Description:

Per https://github.com/pypa/setuptools/issues/1689:

> `setuptools.build_meta:__legacy__` was intended _only_ as a default and was not intended to be specified as the build backend.

Please fix the build system to work correctly with the `setuptools.build_meta`. The explicit use of legacy backend is no longer supported on Gentoo and prevents us from packaging the new gevent version.

mgorny:
Use the non-legacy `setuptools.build_meta` backend.  Prepend the setup directory to `sys.path` prior to importing modules relative to it.

Fixes #1910

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,9 +1,5 @@
 [build-system]
-# Our setup.py imports other things from this directory, meaning
-# in needs to be on sys.path. That's not guaranteed in a PEP517 world;
-# the __legacy__ build module makes that true. Ultimately we need to do that
-# ourself (and/or continue to simplify our build system).
-build-backend = "setuptools.build_meta:__legacy__"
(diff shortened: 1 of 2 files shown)


mgorny:
I'm not 100% sure that this is the best approach but it seems to work here.

jamadden:
Thank you.

mgorny:
Thanks!

PR #1932, opened 2023-02-09
Fix for #1931:  Base socket connect_ex fn should use its own connect implementation.

### LINKED ISSUE #1931: Handshake on SSL wrapped non-blocking sockets results in AttributeError.
* gevent version: 21.12.0 from PyPI
* Python version: Python 3.9.16
* Operating System:  Linux, 4.12.14-coreos, x86_64 GNU/Linux

### Description:

Performing handshake on SSL wrapped non-blocking sockets results in AttributeError.
   
While using gevent monkey patched sockets for creating SSL connections, `do_handshake` function results in an AttributeError. 

```python-traceback
  File "//new.py", line 54, in <module>
    sock.do_handshake()
  File "/usr/local/lib/python3.9/site-packages/gevent/_ssl3.py", line 666, in do_handshake
    self._sslobj.do_handshake()
AttributeError: 'NoneType' object has no attribute 'do_handshake'
```

### Observation:
It seems the `_sslobj` state maintained by `SSLSocket` in `_ssl3.py` is getting reset to `None` here upon connect here: https://github.com/gevent/gevent/blob/master/src/gevent/_ssl3.py#L725

Following is the call trace wrt `SSLSocket` and SocketCommon functions upon calling `connect_ex`:
```
root@priyankar-anc:/usr/local/lib/python3.9/site-packages/gevent# python /new.py
  File "/new.py", line 38, in <module>
    error = sock.connect_ex(address)
  File "/usr/local/lib/python3.9/site-packages/gevent/_ssl3.py", line 738, in connect_ex
    return self._real_connect(addr, True)
  File "/usr/local/lib/python3.9/site-packages/gevent/_ssl3.py", line 715, in _real_connect
    rc = socket.connect_ex(self, addr)
  File "/usr/local/lib/python3.9/site-packages/gevent/_socketcommon.py", line 637, in connect_ex
    return self.connect(address) or 0
  File "/usr/local/lib/python3.9/site-packages/gevent/_ssl3.py", line 733, in connect
    self._real_connect(addr, False)
  File "/usr/local/lib/python3.9/site-packages/gevent/_ssl3.py", line 726, in _real_connect
    traceback.print_stack()
```
For nonblocking sockets the last call to `_real_connect` can raise an exception (corresponding to `EINPROGRESS`) and would reset the `_sslobj` state to `None` .            
 
### What I've run:

```python
from gevent import monkey;
monkey.patch_all()

import errno
import select
import socket
import ssl

IPTOS_PREC_INTERNETCONTROL = 0xc0
DSCP_DEFAULT = IPTOS_PREC_INTERNETCONTROL >> 2
CA_FILE = "/etc/ssl/ca.pem"
CERT_FILE = "/etc/ssl/cert"
PRIV_KEY_FILE = "/etc/ssl/key"
POLLOUT = 0x004

address = ("127.0.0.1", 6651)
sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM, 0)

# Create an SSL context
ctx = ssl.SSLContext(ssl.PROTOCOL_SSLv23)
ctx.verify_mode = ssl.CERT_REQUIRED
ctx.options |= ssl.OP_NO_SSLv2
ctx.options |= ssl.OP_NO_SSLv3
ctx.load_verify_locations(CA_FILE)
ctx.load_cert_chain(CERT_FILE, PRIV_KEY_FILE)
sock = ctx.wrap_socket(sock, do_handshake_on_connect=False)

# Set non-blocking
sock.setblocking(0)
# Set DSCP
sock.setsockopt(socket.IPPROTO_IP, socket.IP_TOS, DSCP_DEFAULT << 2)

# Connect
error = sock.connect_ex(address)
if error not in (0, errno.EINPROGRESS, errno.EWOULDBLOCK):
    sock.close()
    raise Exception("Error connecting to socket")

sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_NODELAY, 1)

# Poll
p = select.poll()
p.register(sock, POLLOUT)

pfds = p.poll(0)

revents = pfds[0][1]
if revents & POLLOUT:
    # Handshake
    sock.do_handshake()
```

priyankar-jain:
This commit attempts to fix #1931 by forcing the Base socket `connect_ex` implementation to use its own `connect` implementation (which handles the exceptions correctly) when called from child class (such as SSLSocket).

### DIFF
--- a/src/gevent/_socketcommon.py
+++ b/src/gevent/_socketcommon.py
@@ -588,15 +588,7 @@ class SocketMixin(object):
         """
         return self.timeout != 0.0
 
-    def connect(self, address):
-        """
-        Connect to *address*.
-
-        .. versionchanged:: 20.6.0
-            If the host part of the address includes an IPv6 scope ID,
-            it will be used instead of ignored, if the platform supplies
-            :func:`socket.inet_pton`.
-        """
+    def __connect(self, address):
         if self.timeout == 0.0:
             return self._sock.connect(address)
         address = _resolve_addr(self._sock, address)
@@ -627,9 +619,20 @@ class SocketMixin(object):
                         result = ECONNREFUSED
                     raise _SocketError(result, strerror(result))
 
+    def connect(self, address):
+        """
+        Connect to *address*.
+
+        .. versionchanged:: 20.6.0
+            If the host part of the address includes an IPv6 scope ID,
+            it will be used instead of ignored, if the platform supplies
+            :func:`socket.inet_pton`.
+        """
+        return self.__connect(address)
+
     def connect_ex(self, address):
         try:
-            return self.connect(address) or 0
+            return self.__connect(address) or 0
         except __socket__.timeout:
             return EAGAIN
         except __socket__.gaierror: # pylint:disable=try-except-raise


jamadden:
Thanks for the report and the PR. I concur with your analysis, based on what the standard library does: `SSLSocket.connect_ex` calls `SSLSocket._real_connect` (just like gevent) and then calls `super().connect_ex()` (just like gevent). Then, `socket.connect_ex` is implemented in C in the standard library,  and calls the same internal C function `internal_connect` that `socket.connect` (also C implementation) does. That's where they differ: gevent's `socket.connect_ex` is implemented in Python to call `self.connect()` --- which of course can be overridden. 

I'm going to tweak the fix slightly and merge.

PR #1944, opened 2023-04-03
Remove the usage of assertRaisesRegexp unit test alias removed in Python 3.12

hrnciar:


jamadden:
Thanks for the PR. This got handled as part of the larger work of supporting Python 3.12.

PR #1950, opened 2023-05-07
Fix asserts in test_attributes

mtelka:
This fixes an obvious typo.  See https://docs.python.org/3/library/socket.html#creating-sockets for reference.


jamadden:
Thanks for the PR. I was able to use a stronger version of this test.

PR #1973, opened 2023-07-17
Added a setter method for the MAX_REQUEST_LINE

Shreesha4994:
WSGI by default limits the url length limit to 8192.
There are scenarios where the user wants to fire URLs exceeding this length.
By creatung a setter method for the constant we can now allow the users to modify the default limit.


jamadden:
Any users who know enough to do something like that can just change the value. No need for an undocumented setter method.

PR #1983, opened 2023-08-04
Fixes gevent.select.select not handling all iterable inputs correctly

### LINKED ISSUE #1979: `gevent.select.select` is too strict on input arguments
* gevent version: 23.7.0 from PyPI
* Python version: 3.10.12 installed through OpenSUSE Tumbleweed
* Operating System: OpenSUSE Tumbleweed

### Description:

This is mostly a nitpick, but it has come up while trying to create type stubs for gevent.

gevent's implementation of `select` mandates that rlist/wlist/xlist are at the very least a `Collection` i.e. they need to be `Iterable`, but they also need to remain unchanged by iteration, since the argument gets passed unmodified to two different functions: the original select implementation (or a  wrapper on windows) and a `SelectResult.select` if the first call didn't already yield a result on one of the file descriptors.

The original implementation accepts an `Iterable`. So technically you would need to first copy each input argument into e.g. a tuple (or split into two iterables using `itertools.tee`, although the truthiness check on win32 would still be incorrect) if the input does not already implement a `__len__`.

I realize this change creates some small overhead, and as such it is not free, for something that should come up rarely, but on the flip side you could end up with some nasty, difficult to track down bugs in external libraries that rely on this part of the specification.

Daverball:
This fixes #1979

### DIFF
--- a/src/gevent/select.py
+++ b/src/gevent/select.py
@@ -159,6 +159,13 @@ def select(rlist, wlist, xlist, timeout=None): # pylint:disable=unused-argument
         # forward compatible
         raise ValueError("timeout must be non-negative")
 
+    # since rlist and wlist can be any iterable we will have to first
+    # copy them into a list, so we can use them in both _original_select
+    # and in SelectResult.select. We don't need to do it for xlist, since
+    # that one will only be passed into _original_select
+    rlist = rlist if isinstance(rlist, (list, tuple)) else list(rlist)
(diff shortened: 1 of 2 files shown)


Daverball:
Not sure why there were multiple appveyor builds for that second commit and why the failing one replaced the successful after a while, but the failure looks like a flaky test to me, unrelated to select. Here's the successful build for comparison: https://ci.appveyor.com/project/denik/gevent/builds/47718618

jamadden:
Thanks!

PR #2004, opened 2023-10-24
Remove useless shebang from testrunner.py

mtelka:
The `testrunner.py` file have Python shebang, but it is not possible to run the file:
```
$ src/gevent/testing/testrunner.py
  File "/tmp/gevent/src/gevent/testing/testrunner.py", line 19, in <module>
    from . import util
ImportError: attempted relative import with no known parent package
$
```
This change simply removes the useless shebang.


mtelka:
Already fixed via https://github.com/gevent/gevent/commit/50dd5a1a70d16899d9065505aa6233ed0bf436c6

PR #2034, opened 2024-05-28
Disable c-ares tests

### LINKED ISSUE #2029: c-ares configure fails with missing `test/Makefile.in`
The c-ares configure could fail with missing `test/Makefile.in`.

This is because:
- the libcares copy in `deps/c-ares` does not contains tests (the `test` directory is missing), and
- building of tests is enabled by default and so when the configure detects all prerequisites for building tests it tries to do so.

Suggested fix:
```
--- gevent-24.2.1/_setupares.py.orig
+++ gevent-24.2.1/_setupares.py
@@ -42,7 +42,7 @@
 ares_configure_command = ' '.join([
     "(cd ", quoted_dep_abspath('c-ares'),
     " && if [ -r include/ares_build.h ]; then cp include/ares_build.h include/ares_build.h.orig; fi ",
-    " && sh ./configure --disable-dependency-tracking -C " + cflags,
+    " && sh ./configure --disable-dependency-tracking --disable-tests -C " + cflags,
     " && cp src/lib/ares_config.h include/ares_build.h \"$OLDPWD\" ",
     " && cat include/ares_build.h ",
     " && if [ -r include/ares_build.h.orig ]; then mv include/ares_build.h.orig include/ares_build.h; fi)",
```

mtelka:
Fixes #2029.

### DIFF
--- a/_setupares.py
+++ b/_setupares.py
@@ -42,7 +42,7 @@ cflags = ('CFLAGS="%s"' % (cflags,)) if cflags else ''
 ares_configure_command = ' '.join([
     "(cd ", quoted_dep_abspath('c-ares'),
     " && if [ -r include/ares_build.h ]; then cp include/ares_build.h include/ares_build.h.orig; fi ",
-    " && sh ./configure --disable-dependency-tracking -C " + cflags,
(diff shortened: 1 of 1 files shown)


jamadden:
Thanks, fixed in another commit.

PR #2070, opened 2024-10-18
Config option to disable report printing in monitoring thread

seagrine:
I do not need the printing and it is generating a lot of spam in my case, but I do need the BlockedEventLoop notification, so I'm proposing a simple config option to disable the printing.

I did not have success with `gevent.get_hub().exception_stream = None`. Maybe it is due to my setup creating multiple hubs/threads beyond my control/awareness. It is hard for me to tell.

### DIFF
--- a/src/gevent/_config.py
+++ b/src/gevent/_config.py
@@ -546,6 +546,19 @@ class MaxBlockingTime(FloatSettingMixin, Setting):
     .. versionadded:: 1.3b1
     """
 
+class PrintReports(BoolSettingMixin, Setting):
+    name = 'print_reports'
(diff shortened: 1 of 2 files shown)


seagrine:
Thanks, Jason! The concern with disabling the exception stream entirely (which we could do via your suggestion or otherwise) is that this is not the only code that uses `exception_stream`, and we want to keep the other usages.

jamadden:
> The concern with disabling the exception stream entirely

Great! I had that reservation too. Hopefully the ability to completely influence what is reported, if at all, in the 22.11 release solves the issue for you.

PR #2077, opened 2024-11-11
Initial compatibility with Cython 3.1a1 by removing legacy Py2 code

### LINKED ISSUE #2076: Not compatible with Cython 3.1 due to Py2 code
Cython 3.1 removes Py2 support, but gevent still has Py2 code in its code base, e.g.

https://github.com/gevent/gevent/blob/07d2b7e4a3c5881e0a3637062b3c7c7c0027971c/src/gevent/libev/corecext.pyx#L64-L69

This leads to build failures for users who have Cython 3.1 installed, as described in

https://github.com/cython/cython/issues/6490

A quick fix would be to exclude Cython 3.1 from the build dependencies:

https://github.com/gevent/gevent/blob/07d2b7e4a3c5881e0a3637062b3c7c7c0027971c/pyproject.toml#L21

However, since Python 2.x support seems no longer intended for gevent, removing the Py2 left-overs seems the best path forward.

jamadden:
Fixes #2076.

@scoder But gevent still won't work, at least on Python 3.12. Cython3.1a1 incorrectly compiles the function `cpdef unsigned int _flags_to_int(object flags) except *:` from `src/gevent/libev/corecext.pyx`. In the case of raising an exception (I've tried a few variations), Cython attempts to do a `Py_DECREF` on a temporary that it already set to NULL. `Py_DECREF` isn't safe to call with a possibly-NULL value, and it crashes.

### DIFF
--- a/src/gevent/greenlet.py
+++ b/src/gevent/greenlet.py
@@ -378,14 +378,8 @@ class Greenlet(greenlet):
         hub = get_my_hub(self) # pylint:disable=undefined-variable
         return hub.loop
 
-    def __nonzero__(self): # pylint:disable=bad-dunder-name
+    def __bool__(self): # pylint:disable=bad-dunder-name
         return self._start_event is not None and self._exc_info is None
-    try:
-        __bool__ = __nonzero__ # Python 3
-    except NameError: # pragma: no cover
-        # When we're compiled with Cython, the __nonzero__ function
-        # goes directly into the slot and can't be accessed by name.
-        pass
 
     ### Lifecycle
 
--- a/src/gevent/libev/corecext.pyx
+++ b/src/gevent/libev/corecext.pyx
@@ -162,7 +162,11 @@ _flags = tuple([
 ])
 
 
-_flags_str2int = dict((string, flag) for (flag, string) in _flags)
+_flags_str2int = {
+    string: flag
+    for flag, string
+    in _flags
+}
 
 
 _events = tuple([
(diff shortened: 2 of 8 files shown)


src/gevent/greenlet.py
```diff
@@ -378,14 +378,8 @@ def loop(self):
         hub = get_my_hub(self) # pylint:disable=undefined-variable
         return hub.loop
 
-    def __nonzero__(self): # pylint:disable=bad-dunder-name
+    def __bool__(self): # pylint:disable=bad-dunder-name
```
scoder, line 381:
```suggestion
    def __bool__(self):
```

src/gevent/libev/corecext.pyx
```diff
@@ -203,32 +202,31 @@ cpdef _flags_to_list(unsigned int flags):
     return result
 
 
-
-basestring = (bytes, str)
-
-
-
-cpdef unsigned int _flags_to_int(object flags) except? -1:
+cpdef unsigned int _flags_to_int(object flags) except *:
```
scoder, line 205: `except? -1` (or whatever integer value you chose) is still more efficient than `except *`. The latter must always ask CPython if something happened, whereas the first only needs to do that if the functions really returned exactly the exception value. I'd recommend sticking with some unlikely exception value here, even arbitrarily chosen. `except *` is mostly useful for `void` or `struct` returns.

scoder:
I created https://github.com/cython/cython/issues/6496 for the Cython code generation issue.

src/gevent/libev/corecext.pyx
jamadden: Thanks. `except *` was one of the things I tried to workaround the code generation issue. I decided to keep it because this isn't performance sensitive code, and there's something vaguely unsettling about comparing an `unsigned int` to -1.

src/gevent/libev/corecext.pyx
```diff
@@ -166,28 +159,34 @@ _flags = [
...
-           (libev.EV_EMBED,    'EMBED'),
-           (libev.EV_FORK,     'FORK'),
-           (libev.EV_CLEANUP,  'CLEANUP'),
-           (libev.EV_ASYNC,    'ASYNC'),
-           (libev.EV_CUSTOM,   'CUSTOM'),
-           (libev.EV_ERROR,    'ERROR')]
+])
+
+
+_flags_str2int = {
```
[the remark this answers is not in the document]
jamadden: Thanks. This isn't performance sensitive, so I'm OK with going through the module right now.

PR #2078, opened 2024-11-11
remove old "from __future__ import ..." statements

a-detiste:


jamadden:
Thank you, but I'm not going to accept this PR. I don't see the need for a change that touches many files and provides to actual benefits.

Issue #2075, opened 2024-11-08
Security: 100-continue causes request smuggling issues


m4yfly:
### Description:

The security issue reported here has not been fixed. https://huntr.com/bounties/1739c8fc-e112-49f5-a006-cd2203746172

jamadden:
Don't use pywsgi in a security conscious environment. [It is clearly documented](http://www.gevent.org/api/gevent.pywsgi.html) that pywsgi is not meant for production or exposure to the Internet.

That said, tests and patches are certainly welcome.

muhammedhunaid:
@m4yfly , is the cause of this issue causing all previous builds of gevent to break?

Problem: both of the following versions of gevent broke: 23.7.0 (Run 1; ideal setup) and 21.12.0 ([Run 2; online workaround](https://github.com/gevent/gevent/issues/1899#issue-1330048647)) 
Environment: cython==3.0.0.11 (Run 1), cython==3.0.0a10 (Run 2), and ARM Silicon (M1)

m4yfly:
@muhammedhunaid No, this issue is not related to stability; it's only about information leakage.

jamadden:
Thank you. Fix merged in https://github.com/gevent/gevent/commit/cbb527d3096502ae251f83002e4a4c0c024c18a9

PR #2079, opened 2024-11-12
Fix for #2075: Resolved the request smuggling issue caused by the handling of 100-continue.

m4yfly:
Added new test cases; used the `send_100_continue_enabled` flag to fix the logic in the `_discard` function, ensuring that the `send_100_continue` is not triggered during the `_discard` process.


jamadden:
Thank you. Fix merged in https://github.com/gevent/gevent/commit/cbb527d3096502ae251f83002e4a4c0c024c18a9

PR #2083, opened 2024-12-04
Fix backdoor support for unix socket

### LINKED ISSUE #1769: BackdoorServer doesn't work anymore with AF_UNIX listeners
* gevent version: 20.12.1 from pypi
* Python version: 2.7.18 from the official Fedora repos.
* Operating System: Fedora 5.8.10-200.fc32.x86_64

### Description:

`gevent.backdoor.BackdoorServer()`  doesn't work when given an `AF_UNIX` listener. It assumes the listener is `AF_INET` and tries to do a `conn.setsockopt(socket.SOL_TCP, socket.TCP_NODELAY, True)` in `BackdoorServer.handle()`.

Previously, on an older gevent version unix sockets worked fine. In some cases they're preferable over TCP on localhost, because basic authentication can be done with the socket file owner/permissions.

We can probably check the socket type in `handle()` with `getsockopt(SO_DOMAIN)` or something. Thoughts?

oz123:
This is an attempt to fix #1769.

@danmilon.

### DIFF
--- a/docs/changes/1769.bugfix.rst
+++ b/docs/changes/1769.bugfix.rst
@@ -1,0 +1,1 @@
+Fix using gevent's BackdoorServer with Unix sockets.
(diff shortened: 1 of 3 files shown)


jamadden:
Thanks!

oz123:
You're welcome!

PR #2103, opened 2025-04-22
Deal with BrokenPipeError since openssl 3.4.x.

pmgdeb:
Follows from the changes in gh-127257 and the test adaption in gh-115627.

[gh-127257: ssl: Raise OSError for ERR_LIB_SYS](https://github.com/python/cpython/pull/127361)
[gh-115627: Fix ssl test_pha_required_nocert()](https://github.com/python/cpython/pull/117821)

We are building openSUSE Factory with OpenSSL 3.5 and we can see that gevent fails the test__server.py with BrokenPipeError, here is the error output:

```
[  147s]   ..sTraceback (most recent call last):
[  147s]     File "src/gevent/greenlet.py", line 906, in gevent._gevent_cgreenlet.Greenlet.run
[  147s]       result = self._run(*self.args, **self.kwargs)
[  147s]     File "/home/abuild/rpmbuild/BUILD/python-gevent-24.10.3-build/BUILDROOT/usr/lib64/python3.11/site-packages/gevent/baseserver.py", line 34, in _handle_and_close_when_done
[  147s]       return handle(*args_tuple)
[  147s]              ^^^^^^^^^^^^^^^^^^^
[  147s]     File "/home/abuild/rpmbuild/BUILD/python-gevent-24.10.3-build/BUILDROOT/usr/lib64/python3.11/site-packages/gevent/server.py", line 210, in wrap_socket_and_handle
[  147s]       return self.handle(ssl_socket, address)
[  147s]              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[  147s]     File "/home/abuild/rpmbuild/BUILD/python-gevent-24.10.3-build/BUILDROOT/usr/lib64/python3.11/site-packages/gevent/tests/test__server.py", line 27, in handle
[  147s]       request_line = fd.readline()
[  147s]                      ^^^^^^^^^^^^^
[  147s]     File "/usr/lib64/python3.11/socket.py", line 718, in readinto
[  147s]       return self._sock.recv_into(b)
[  147s]              ^^^^^^^^^^^^^^^^^^^^^^^
[  147s]     File "/home/abuild/rpmbuild/BUILD/python-gevent-24.10.3-build/BUILDROOT/usr/lib64/python3.11/site-packages/gevent/ssl.py", line 635, in recv_into
[  147s]       return self.read(nbytes, buffer)
[  147s]              ^^^^^^^^^^^^^^^^^^^^^^^^^
[  147s]     File "/home/abuild/rpmbuild/BUILD/python-gevent-24.10.3-build/BUILDROOT/usr/lib64/python3.11/site-packages/gevent/ssl.py", line 437, in read
[  147s]       bytes_read += self._sslobj.read(nbytes, buffer)
[  147s]                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[  147s]   BrokenPipeError: [Errno 32] Broken pipe
[  147s]   2025-04-20T23:39:43Z <Greenlet at 0x7f5d518b5d90: _handle_and_close_when_done(<bound method StreamServer.wrap_socket_and_handle , <bound method StreamServer.do_close of <SimpleStre, (<gevent._socket3.socket [closed] at 0x7f5d50cfae5)> failed with BrokenPipeError

```

Please, let me know if you need more info. TIA

### DIFF
--- a/src/greentest/3.10/test_ssl.py
+++ b/src/greentest/3.10/test_ssl.py
@@ -2485,7 +2485,6 @@ class ThreadedEchoServer(threading.Thread):
                 # See also http://erickt.github.io/blog/2014/11/19/adventures-in-debugging-a-potential-osx-kernel-bug/
                 if e.errno != errno.EPROTOTYPE and sys.platform != "darwin":
                     self.running = False
-                    self.server.stop()
                     self.close()
                 return False
             else:
@@ -2620,9 +2619,6 @@ class ThreadedEchoServer(threading.Thread):
                     self.close()
                     self.running = False
 
-                    # normally, we'd just stop here, but for the test
-                    # harness, we want to stop the server
-                    self.server.stop()
 
     def __init__(self, certificate=None, ssl_version=None,
                  certreqs=None, cacerts=None,
@@ -2657,21 +2653,33 @@ class ThreadedEchoServer(threading.Thread):
         self.conn_errors = []
         threading.Thread.__init__(self)
         self.daemon = True
+        self._in_context = False
 
     def __enter__(self):
+        if self._in_context:
+            raise ValueError('Re-entering ThreadedEchoServer context')
+        self._in_context = True
         self.start(threading.Event())
         self.flag.wait()
         return self
 
     def __exit__(self, *args):
+        assert self._in_context
+        self._in_context = False
         self.stop()
         self.join()
 
     def start(self, flag=None):
+        if not self._in_context:
+            raise ValueError(
+                'ThreadedEchoServer must be used as a context manager')
         self.flag = flag
         threading.Thread.start(self)
 
     def run(self):
+        if not self._in_context:
+            raise ValueError(
+                'ThreadedEchoServer must be used as a context manager')
         self.sock.settimeout(1.0)
         self.sock.listen(5)
         self.active = True
--- a/src/greentest/3.11/test_ssl.py
+++ b/src/greentest/3.11/test_ssl.py
@@ -2492,7 +2492,6 @@ class ThreadedEchoServer(threading.Thread):
                 # See also http://erickt.github.io/blog/2014/11/19/adventures-in-debugging-a-potential-osx-kernel-bug/
                 if e.errno != errno.EPROTOTYPE and sys.platform != "darwin":
                     self.running = False
-                    self.server.stop()
                     self.close()
                 return False
             else:
@@ -2627,10 +2626,6 @@ class ThreadedEchoServer(threading.Thread):
                     self.close()
                     self.running = False
 
-                    # normally, we'd just stop here, but for the test
-                    # harness, we want to stop the server
-                    self.server.stop()
-
     def __init__(self, certificate=None, ssl_version=None,
                  certreqs=None, cacerts=None,
                  chatty=True, connectionchatty=False, starttls_server=False,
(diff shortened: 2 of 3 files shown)


jamadden:
I am unable to reproduce this on Python 3.13.3 with OpenSSL 3.5.0 on either macOS or Linux (the manylinux 2014 image, in which I manually upgraded OpenSSL). Which specific test is failing, and how is it failing? (i.e., what's the rest of the log?)

That aside, it's not clear to me that the approach in this PR is correct. EPIPE/BrokenPipeError means the remote connection has dropped unexpectedly; this is a layer below the SSL framing layer. Typically, you want this exception to propagate so the user of the socket knows there's been a problem; the standard library didn't add any code to hide this error when it occurs.

pmgdeb:
Hi, it fails in python311 but not for newer python versions. Here is a [good build](https://build.opensuse.org/package/live_build_log/openSUSE:Factory/python-gevent/standard/x86_64) and the attached file contains the build failure with OpenSSL 3.5.0. 
[gevent_log.txt](https://github.com/user-attachments/files/19887341/gevent_log.txt)

Thanks for you answer, I also think this should be handled only in the regression test and not in src/gevent/ssl.py to correctly propagate the exception. I'll change the PR as soon as possible.

jamadden:
The logs you provided don't show an error in gevent's `test__server.py`. Instead, they show several errors in the standard library's `test_ssl.py`, all of type `ConnectionRefusedError`.

I can indeed reproduce that failure using Python 3.11.12 and OpenSSL 3.5.0. (And I can't produce a failure in `test__server.py`).

*However*, I reproduce that failure whether or not I use gevent. That is, `python -m gevent.testing.monkey_test test_ssl.py` (to run the test with gevent) and `python -m test.test_ssl` (to run the unchanged standard library test) produce the same `ConnectionRefusedError`. 

This indicates it is not a problem in gevent. Instead, it is a problem using Python 3.11.12 (which is only tested with OpenSSL 3.0) with OpenSSL 3.5.

jamadden:
The relevant CPython change appears to be [this one](https://github.com/python/cpython/issues/126500), which appears to only have been backported to 3.12, not 3.11 or older, probably because those versions are either in security-fix-only mode or unsupported altogether.

pmgdeb:
I think gevent is running a copy of the test_ssl.py from cpython but its missing this recent commit: [gh-126500: test_ssl: Don't stop ThreadedEchoServer on OSError in ConnectionHandler;](https://github.com/python/cpython/pull/126503). This cpython PR was ported to [3.12](https://github.com/python/cpython/pull/126572) and [3.13](https://github.com/python/cpython/pull/126571/files) but not for 3.11.

I have verified that adding those changes fixes it for 3.12 and 3.13 as well with the latest version. Note that, it was failing for us with 3.11 as we ported some of the fixes in 3.12 to work with OpenSSL 3.5.0.

I think you would need to update the test_ssl to sync with cpython, I can change this PR to add those changes. Please, let me know your view on this. TIA

jamadden:
We don't make changes to the standard library tests. We don't backport changes the coredevs don't.

jamadden:
No, we can't just use the tests from some OS provided package. They're an incredibly important part of gevent's test suite and as such must always be available no matter what platform we're running on. I don't have time to deal with figuring out what package needs installed on what OS, figuring out how to do that, writing instructions, dealing with non-root users, etc, etc, etc. 

Closing this as wontfix.

Issue #2111, opened 2025-05-29
Exception ignored, AssertionError


he1pu:
### des
I encountered this exception when running the following code.
log:
```
Exception ignored in: <bound method _ForkHooks.after_fork_in_child of <gevent.threading._ForkHooks object at 0x7fbb9733a410>>
Traceback (most recent call last):
  File "/var/www/ipasign/myvenv/lib/python3.11/site-packages/gevent/threading.py", line 398, in after_fork_in_child
    assert not thread.is_alive()
               ^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.11/threading.py", line 1437, in is_alive
    assert not self._is_stopped and self._started.is_set()
AssertionError: 
out:  hello world
err:  
[2025-05-29 02:34:10] app INFO: hello world
```
Command:`gunicorn -w 4 -k gevent -b 0.0.0.0:5500 testapp:app`

#### ps
+ 1: When I add:
 ```
from gevent import monkey
monkey.patch_all()
``` 
in the header of `apptest.py`, this exception still occurs.
+ 2: When I delete: `logger.info(' background background background')`, the exception no longer occurs

I hope to receive some help. Thank you

### codes

`testapp.py:`
```
# testapp.py
import gevent
from flask import Flask, jsonify
from testz import testdef
from mylog import logger

# Flask app init
app = Flask(__name__)

def work():
    result = testdef()
    logger.info(result)

@app.route('/test/')
def indextest():
    logger.info(' background background background')
    gevent.spawn(work)
    return jsonify({"message": "start"}), 200

if __name__ == '__main__':
    app.run(host='0.0.0.0', port=5500)

```

`testz.py :`
```
#testz.py
import gevent.subprocess as gsubprocess
from copy import copy

def testdef():
    process = gsubprocess.run(['echo', 'hello world'], capture_output=True)
    stdout_str = process.stdout.decode('utf-8', errors='ignore').strip() if process.stdout else ""
    stderr_str = process.stderr.decode('utf-8', errors='ignore').strip() if process.stderr else ""
    print('out: ', stdout_str)
    print('err: ', stderr_str)
    return copy(stdout_str)

```

`mylog.py:`
```
import os
import logging
from logging.handlers import RotatingFileHandler
from dotenv import load_dotenv

ENV_PATH = os.getenv('ENV_PATH')
if ENV_PATH and os.path.exists(ENV_PATH):
    load_dotenv(ENV_PATH)
else:
    load_dotenv()

[...]

PR #2115, opened 2025-06-06
Remove the (not thread.is_alive()) assertion

vbraun:
This assertion apparently has a small chance of failing, possibly by a race condition. It was already skipped on Python 3.13+.

Fixes https://github.com/gevent/gevent/issues/2111


vbraun:
Obsoleted by https://github.com/gevent/gevent/pull/2117

PR #2129, opened 2025-08-17
Restore function of signal.set_wakeup_fd for SIGCHLD

### LINKED ISSUE #2126: signal.set_wakeup_fd doesn't appear to be functional wrt SIGCHLD after monkeypatching
Multiprocessing forkserver relies on the following code:

https://github.com/python/cpython/blob/c5cebe1b5afa58fc0ee95153a4b1905229dce7dc/Lib/multiprocessing/forkserver.py#L231-L249

and 

https://github.com/python/cpython/blob/c5cebe1b5afa58fc0ee95153a4b1905229dce7dc/Lib/multiprocessing/forkserver.py#L274-L276

After monkeypatching the `sig_w` is never written to. Adding the allegedly racy write to the `sigchld_handler` restores functionality.

arcivanov:
The cause of this issue is that when no SIGCHLD handler is set set_wakeup_fd does not function at all. In Gevent SIGCHLD handler is never set (unless SIG_IGN or SIG_DFL).

This change emulates the CPython's set_wakeup_fd behavior using signal._on_child_hook signal handler.

fixes #2126


arcivanov:
@jamadden please review

PR #2187, opened 2026-07-13
Fix bare except in _addresses.py: use except struct.error

koteshyelamati:
The bare `except:` clause in `_ipv4_inet_aton()` catches `BaseException`, including `SystemExit` and `KeyboardInterrupt`. The only exception that can realistically be raised here is `struct.error` (when an integer part exceeds the 0–255 byte range for `struct.pack('BBBB', ...)`), since `int(part)` on valid digit bytes succeeds normally.

Changing `except:` to `except struct.error:` makes the intent explicit and prevents accidentally silencing signals or shutdown exceptions.

### DIFF
--- a/src/gevent/resolver/_addresses.py
+++ b/src/gevent/resolver/_addresses.py
@@ -60,7 +60,7 @@ def _ipv4_inet_aton(text):
     try:
         ints = [int(part) for part in parts]
         return struct.pack('BBBB', *ints)
-    except:
+    except struct.error:
         raise AddressSyntaxError(text)
 
 


src/gevent/resolver/_addresses.py
```diff
@@ -60,7 +60,7 @@ def _ipv4_inet_aton(text):
     try:
         ints = [int(part) for part in parts]
         return struct.pack('BBBB', *ints)
-    except:
+    except struct.error:
```
jamadden, line 63: It's totally possible for this to raise a `ValueError` too, from `int(part)`, if `part` is long enough, so `struct.error` is not enough. Rather than list them both, and in the interest of keeping behaviour backward compatible (for cases I haven't thought of --- MemoryError is another possibility), I'd prefer `except Exception:`. That solves the issues you bring up which were all `BaseException`.


```suggestion
    except Exception:
        # Used to catch `BaseException`. 
        # We expect struct.error or ValueError,
        # but historically we've caught everything, so 
        # we're mostly leaving that for BWC (e.g, what if it raises
        # MemoryError), just no longer
        # catching BaseException.
```

jamadden (changes requested):
Thanks! I want to take a slightly different tack though.

PR #2189, opened 2026-07-17
Run the threading atexit hooks before joining non-daemon threads

### LINKED ISSUE #2188: A non-daemon thread that waits on an atexit hook never exits.
* Operating System: (Ubuntu 22.04.5 LTS 6.8.0-111-generic x86_64)

### Description:

This worked on 3.12, but broke in 3.13+. I'll try to make a pull request for it.

### What I've run:

```python
❯ uv run --python=3.14.6 --script repro_gevent_shutdown_deadlock.py
Installed 4 packages in 7ms
HANGS: gevent joins the thread before releasing it (child exit 1)

❯ uv run --python=3.12 --script repro_gevent_shutdown_deadlock.py
exits cleanly: gevent now runs the atexit hooks before joining. Drop the workaround.

```

```python
# /// script
# dependencies = ["gevent==26.5.0", "greenlet==3.5.3"]
# ///

import subprocess
import sys


def child():
    from gevent import monkey

    monkey.patch_all()

    import faulthandler
    import threading

    faulthandler.dump_traceback_later(5, exit=True)  # exit 1 if still alive then

    done = threading.Event()
    threading._register_atexit(done.set)  # type: ignore[missing-attribute]
    threading.Thread(target=done.wait).start()  # non-daemon, waits for that hook


def main():
    proc = subprocess.run([sys.executable, __file__, "child"], capture_output=True, timeout=60)
    if proc.returncode == 0:
        print("exits cleanly: gevent now runs the atexit hooks before joining. Drop the workaround.")
        return 0
    print(f"HANGS: gevent joins the thread before releasing it (child exit {proc.returncode})")
    return 1


if __name__ == "__main__":
    if len(sys.argv) > 1:
        child()
    else:
        sys.exit(main())

```

ddorian:
fixes #2188


jamadden:
Thanks. I've confirmed the fix. Again there are test failures, but not the fault of this PR which I'll deal with separately.

PR #2191, opened 2026-07-24
Replace concurrent.futures' _global_shutdown_lock when patching threads

### LINKED ISSUE #1865: Popen stall in ThreadPoolEecutor under docker

ddorian:
Executor.submit holds it across Thread.start(), so a worker greenlet that forks runs the os.register_at_fork handlers it is registered with while another greenlet holds it. A native lock deadlocks the process; a cooperative one parks the forking greenlet inside os.fork(), letting a second greenlet fork before the first has finished.

Fixes #1865.

### DIFF
--- a/docs/changes/1865.bugfix.rst
+++ b/docs/changes/1865.bugfix.rst
@@ -1,0 +1,9 @@
+Replace ``concurrent.futures.thread._global_shutdown_lock`` when patching
+threads, importing that module if needed. ``Executor.submit`` holds it across
+``Thread.start()``, so a worker greenlet that forks runs the
+:func:`os.register_at_fork` handlers it is registered with while another
+greenlet holds it: a native lock deadlocked, a cooperative one parked the
+greenlet inside ``os.fork()`` (which ``filelock`` 3.30 rejects).
+
+Like the rest of ``patch_thread(existing_locks=True)``, this needs the process
+to be single threaded when patching.
--- a/src/gevent/monkey/__init__.py
+++ b/src/gevent/monkey/__init__.py
@@ -352,6 +352,9 @@ def patch_thread(threading=True, _threading_local=True, Event=True, logging=True
         best-effort attempt and, on certain implementations, may not detect all
         locks. It is important to monkey-patch extremely early in the startup process.
         Setting this to False is not recommended, especially on Python 2.
+        This also replaces the lock :mod:`concurrent.futures.thread` registers
+        with :func:`os.register_at_fork`, which deadlocks or switches greenlets
+        mid-fork.
 
     .. caution::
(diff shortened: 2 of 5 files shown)


ddorian:
Create these to see the test failures in action:
https://github.com/gevent/gevent/pull/2193
https://github.com/gevent/gevent/pull/2192

jamadden:
Thanks!

PR #2192, opened 2026-07-24
Add a failing test for greenlet switches inside os.fork()

ddorian:
Reproduces the other half of gh-1865 without the fix: importing concurrent.futures.thread after patch_all() gives it a cooperative lock, and acquiring that lock parks the greenlet mid-fork, so a second fork begins.


jamadden:
Only added as an example for #2191

PR #2193, opened 2026-07-24
Add a failing test for the concurrent.futures fork deadlock

ddorian:
Reproduces gh-1865 without the fix: importing concurrent.futures.thread before patch_all() leaves a native lock registered with os.register_at_fork, and a greenlet that forks while another holds it hangs the process.


jamadden:
Only added as an example for #2191

PR #2195, opened 2026-07-29
Stop Popen.__exit__ raising 'reentrant call' on a pipe a greenlet is reading

### LINKED ISSUE #2194: gevent: ``Popen.__exit__`` closes a pipe another greenlet is still reading, and raises ``RuntimeError: reentrant call`` out of the cleanup path
### Description:

``communicate()`` spawns a greenlet per pipe. Leave the ``with`` block while one is still parked and ``__exit__`` closes that pipe out from under it; greenlets share a thread, so CPython raises instead of blocking on the buffered object's lock.
``communicate()`` guards the identical close and ``__exit__`` does not, and the raise lands before its ``try/finally``, so the ``wait()`` that reaps the child is skipped too.

```python
❯ uv run gevent_popen_exit_reentrant_repro.py
gevent 26.7.0 | python 3.14.6
REPRODUCED: reentrant call inside <_io.BufferedReader name=6>
child 807861 was left unreaped: __exit__ raised before its wait()
```

### What I've run:

```python
from gevent import monkey

monkey.patch_all()

import sys  # noqa: E402

import gevent  # noqa: E402
from gevent.subprocess import PIPE, Popen  # noqa: E402

print(f"gevent {gevent.__version__} | python {sys.version.split()[0]}")

try:
    # A child that neither writes nor exits, so the reader stays parked.
    with Popen([sys.executable, "-c", "import time; time.sleep(3600)"],
               stdout=PIPE) as popen:
        gevent.spawn(popen.communicate)  # parks a greenlet inside popen.stdout
        gevent.sleep(0.5)
        popen.kill()  # so that __exit__'s wait() has a dead child to reap
except RuntimeError as ex:
    print(f"REPRODUCED: {ex}")
    print(f"child {popen.pid} was left unreaped: __exit__ raised before its wait()")
    sys.exit(0)

print("NOT REPRODUCED: __exit__ closed the pipe without raising.")
sys.exit(1)
```

ddorian:
fixes #2194 

See failing tests:

https://github.com/ddorian/gevent/pull/1

https://github.com/ddorian/gevent/pull/2

### DIFF
--- a/docs/changes/2194.bugfix.rst
+++ b/docs/changes/2194.bugfix.rst
@@ -1,0 +1,8 @@
+Stop the greenlets that ``communicate()`` spawned before
+``gevent.subprocess.Popen.__exit__`` closes the child's pipes. Leaving the
+``with`` block while one of them was still parked in a pipe, because an
+exception was propagating or because the greenlet running the block was
+killed, raised ``RuntimeError: reentrant call`` out of ``__exit__``. That
+replaced the exception that was really unwinding, and skipped the ``wait()``
+that reaps the child. A pipe some other greenlet is reading is now left alone
+rather than raising, which is what ``communicate()`` already did.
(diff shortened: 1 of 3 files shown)


jamadden:
Thanks!

PR #2199, opened 2026-08-07
Fix semaphore notifications from hubless native threads

### LINKED ISSUE #2013: Logging lock not released when held by OS threads
* gevent version: 23.9.1, installed from PyPi via Poetry
* Python version: cPython 3.9.18 installed via pyenv
* Operating System: OSX Sonoma 14.2

### Description:

In a monkey-patched environment, I'm seeing that when the main thread attempts to acquire a logging lock held by an OS thread, it will wait long after the OS thread releases the lock. We have observed this waiting up to 5 hours before successfully acquiring the lock.

I believe this is a complete repro case:

```python
import logging
import time
import gevent
import gevent.monkey

gevent.monkey.patch_all()

pool = gevent.get_hub().threadpool

class PausingHandler(logging.StreamHandler):
    def emit(self, record):
        time.sleep(1)
        return super(PausingHandler, self).emit(record)
    
logging.basicConfig(level=logging.INFO, handlers=[PausingHandler()], format="%(message)s")

logger = logging.getLogger("gevent_test")

def log_async(msg):
    gevent.get_hub()
    while True:
        logger.info(msg)
        time.sleep(1)

def main():
    thread = pool.apply_async(log_async, ["IN_THREAD"])
    greenlet = gevent.spawn(log_async, "IN_GREENLET")
    
    thread.join()
    
if __name__ == "__main__":
    main()
```

Which produces this output:
```
IN_GREENLET
IN_THREAD
IN_THREAD
IN_THREAD
...
```

Following the suggestion in https://www.gevent.org/changelog.html#id30, I added `gevent.get_hub()` to the top of `log_async`, which seems to have fixed the problem.

Thanks for all your hard work on gevent! Its an amazing achievement.

florentinl:
Fixes #2013.

`dd-trace-py` encountered this through Python logging. After monkey-patching, logging handler locks are backed by gevent semaphores, so a hubless native thread logging alongside greenlets could release a handler lock without waking the waiting greenlets.

A native thread can release a gevent semaphore while a greenlet is waiting for it. The semaphore becomes available, but its notification was attempted on the native thread. Greenlets cannot switch across threads, so the waiter could remain parked indefinitely.

When this happens, schedule the normal notification on the semaphore's existing owning hub with `run_callback_threadsafe`. This avoids the cross-thread switch while preserving the existing notification and acquisition order. The normal same-hub path is unchanged.

### DIFF
--- a/src/gevent/_abstract_linkable.py
+++ b/src/gevent/_abstract_linkable.py
@@ -220,11 +220,20 @@ class AbstractLinkable(object):
             try:
                 hub = self._capture_hub(False) # Must create, we need it.
             except InvalidThreadUseError:
-                # The current hub doesn't match self.hub. That's OK,
-                # we still want to start the notifier in the thread running
-                # self.hub (because the links probably contains greenlet.switch
-                # calls valid only in that hub)
-                pass
+                # Links are normally greenlet.switch methods and must be
+                # notified by their owning hub. Wake that hub without first
+                # attempting an unsafe switch from this native thread.
+                hub = self.hub
+                loop = hub.loop
+                if loop is None:
+                    # The owner was destroyed while this thread was using the
+                    # linkable. Its callbacks cannot safely run anywhere else.
+                    raise
+                self._notifier = loop.run_callback_threadsafe(
+                    self._notify_links,
+                    []
+                )
+                return
             if hub is not None:
                 self._notifier = hub.loop.run_callback(self._notify_links, [])
             else:
--- a/src/gevent/_abstract_linkable.py
+++ b/src/gevent/_abstract_linkable.py
@@ -243,6 +243,14 @@ class AbstractLinkable(object):
                     # The owner was destroyed while this thread was using the
                     # linkable. Its callbacks cannot safely run anywhere else.
                     raise
+                # The check on self._notifier above and this assignment are
+                # not atomic: another thread could install its own notifier
+                # in between, and one of the two would be lost. That's safe
+                # per the class comment above (2a/2b) when this is compiled
+                # with Cython (the GIL is held for the whole method) or the
+                # subclass holds a native lock (only Semaphore does). In
+                # pure-Python mode without such a lock (Event, AsyncResult),
+                # this is a real, currently-unclosed race.
                 notifier = _FakeNotifier()
                 self._notifier = notifier
                 try:
--- a/src/gevent/_abstract_linkable.py
+++ b/src/gevent/_abstract_linkable.py
@@ -37,13 +37,24 @@ __all__ = [
 _get_thread_ident = __import__(thread_mod_name).get_ident
 _allocate_thread_lock = __import__(thread_mod_name).allocate_lock
 
+# Reserve _notifier before a foreign thread wakes the owner loop. It mirrors
+# the callback attributes used by the wait and unlink paths.
 class _FakeNotifier(object):
     __slots__ = (
+        'args',
         'pending',
     )
 
     def __init__(self):
+        self.args = ([],)
+        self.pending = True
+
+    def stop(self):
+        self.args = None
         self.pending = False
+
+    def __bool__(self):
+        return self.args is not None
 
 def get_roots_and_hubs():
     from gevent.hub import Hub # delay import
@@ -221,18 +232,28 @@ class AbstractLinkable(object):
                 hub = self._capture_hub(False) # Must create, we need it.
             except InvalidThreadUseError:
                 # Links are normally greenlet.switch methods and must be
-                # notified by their owning hub. Wake that hub without first
-                # attempting an unsafe switch from this native thread.
+                # notified by their owning hub. First publish a notifier that
+                # coalesces foreign-thread wakeups, then wake that hub. The
+                # callback returned by run_callback_threadsafe cannot itself
+                # be used as _notifier: the hub may run it before this thread
+                # has received and stored it.
                 hub = self.hub
                 loop = hub.loop
                 if loop is None:
                     # The owner was destroyed while this thread was using the
                     # linkable. Its callbacks cannot safely run anywhere else.
                     raise
-                self._notifier = loop.run_callback_threadsafe(
-                    self._notify_links,
-                    []
-                )
+                notifier = _FakeNotifier()
+                self._notifier = notifier
+                try:
+                    loop.run_callback_threadsafe(
+                        self._notify_links_from_threadsafe,
+                        notifier
+                    )
+                except:
+                    if self._notifier is notifier:
+                        self._notifier = None
+                    raise
                 return
             if hub is not None:
                 self._notifier = hub.loop.run_callback(self._notify_links, [])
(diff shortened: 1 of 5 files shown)


jamadden:
Unfortunately as you can see, this breaks several other tests and assertions.

florentinl:
Yes thank you for the CI approval. I see how to replicate it locally. Sorry for bothering you

florentinl:
I was introducing a race condition error by running `self._notifier = run_callback_threadsafe(self._notify_links, [])` as the `_notify_links` callback could run before self._notifier was set and go down the wrong `self._notifier is None` path.

I got around it by assigning the _notifier to a `_FakeNotifier` and doing a notifier identity check to ensure that the _notifier is not stale.

It should pass the CI now. Hopefully...

src/gevent/_abstract_linkable.py
```diff
@@ -236,6 +266,16 @@ def _check_and_notify(self):
                 finally:
                     self._notifier = None
 
+    def _notify_links_from_threadsafe(self, notifier):
+        # The callback may be stale after cancellation, fork, or a newer
+        # notification. Only the notifier that requested this wakeup may run.
+        if self._notifier is notifier:
+            if notifier:
+                notifier.pending = False
+                self._notify_links(notifier.args[0])
```
jamadden, line 292: The only time `_notify_links_from_threadsafe` is invoked, it is passed a `_FakeNotifier`, constructed with just `_FakeNotifier()`, meaning it has no `args` -- and the `args` property is never mutated by this code. So how would this branch ever run and do something useful? (What am I missing/forgetting?)

src/gevent/_abstract_linkable.py
```diff
@@ -220,11 +231,30 @@ def _check_and_notify(self):
...
+                notifier = _FakeNotifier()
+                self._notifier = notifier
+                try:
+                    loop.run_callback_threadsafe(
+                        self._notify_links_from_threadsafe,
+                        notifier
+                    )
+                except:
+                    if self._notifier is notifier:
+                        self._notifier = None
```
jamadden, line 263: This section is full of data races, is it not, if we're running in pure-Python mode or without the GIL? i.e., the notifier could be changed between the check and the assignment, invalidating the check. This should be noted in a comment.

Getting cross-thread operations fully right is going to require introducing some native locks.

I won't ask you to do that, that's a bigger architectural change and this may be a sufficient improvement on its own.

src/gevent/_gevent_c_abstract_linkable.pxd
```diff
@@ -57,6 +58,7 @@ cdef class AbstractLinkable(object):
    cpdef unlink(self, callback)
 
    cdef _check_and_notify(self)
+   cpdef _notify_links_from_threadsafe(self, notifier)
```
jamadden: Why does this need to be `cpdef` and not just `cdef`?

src/gevent/_gevent_c_abstract_linkable.pxd
```diff
@@ -36,7 +36,8 @@ cdef void _init()
 cdef dict get_roots_and_hubs()
 
 cdef class _FakeNotifier(object):
-    cdef bint pending
+    cdef public object args
+    cdef public bint pending
```
jamadden, line 40: Why do these need to be public? (Exposed as writable properties to Python when we're running with compiled code; only other compiled code should be using these objects in that case).

src/gevent/tests/test__event.py
```diff
@@ -279,6 +279,72 @@ def test_cross_thread_use_set_in_bg2(self):
         # Do it again to make sure it works multiple times.
         self.test_cross_thread_use_set_in_bg()
 
+    @greentest.ignores_leakcheck
+    def test_cross_thread_callback_can_run_before_scheduling_returns(self):
+        loop = gevent.get_hub().loop
+        if not hasattr(loop, '__dict__'):
+            self.skipTest("The loop's methods cannot be wrapped")
```
jamadden: Loops only have a `__dict__` when we're testing with the FFI-based implementations, which long-term will be going away. 

Have you tried subclassing? If the Cython `run_callback_threadsafe` is declared `cpdef` you can override or replace it in a subclass: 
```python
import gevent
class Loop(type(gevent.get_hub().loop):
    def run_callback_threadsafe(...):
        super().run_callback_threadsafe()

gevent.get_hub().loop = Loop()
```

(Untested, may or may not work as desired.)

src/gevent/_abstract_linkable.py
florentinl: The assumption is that if the object is already in its ready state (a foreign thread already called `set()/release()`) but the scheduled callback hasn't run yet, and a greenlet calls `wait()` during that gap, it appends its resume-callback into the pending notifier's `args[0]` instead of registering a normal rawlink: https://github.com/gevent/gevent/blob/master/src/gevent/_abstract_linkable.py#L525

src/gevent/_abstract_linkable.py
florentinl: Agreed, without the GIL or in pure-python mode it is not very sound. I'll add a comment. Added comment in ce7996d7

src/gevent/_gevent_c_abstract_linkable.pxd
florentinl: It desn't need to. That's just me not being very familiar with cython. Thank you.
Fixed in ce7996d7

src/gevent/_gevent_c_abstract_linkable.pxd
florentinl: I must admit, I just added both to silence an `AttributeError: '_FakeNotifier' object has no attribute 'pending'.`

After looking a bit more into it, it looks like the fact that `AbstractLinkable._notifier` is an untyped slot means that when it is accessed without knowing the concrete type it uses python-level attribute protocol which is only generated with public.

It should be avoidable though. I'll look into it, thank you.

src/gevent/_abstract_linkable.py
jamadden: Right, ok. Probably worth a comment to that effect.

src/gevent/_gevent_c_abstract_linkable.pxd
florentinl: I can't really find an easy way to make it non public. I think it could make sense to keep it public as it is already the case with counter part `callback` used in the same thread case: https://github.com/gevent/gevent/blob/8bbfe7c2c50d4b18e562a0a336df888649a88fb3/src/gevent/libev/corecext.pyx#L287

src/gevent/_abstract_linkable.py
florentinl: Added a comment in d4f5f098

src/gevent/tests/test__event.py
florentinl: Good suggestion, I moved the tests to this approach and it looks good. Now all backends are tested. Fixed in 9b049915

jamadden:
Thanks!

PR #2200, opened 2026-08-14
gevent/subprocess.py: give the pre-exec child a hub of its own

### LINKED ISSUE #2196: spawning a subprocess poisons the process's TLS connections
### Description:

An at-fork child handler that yields (OpenTelemetry, Sentry and filelock all register one) runs inside fork() and hands control to the child's copy of the hub, which resumes copies of the parent's greenlets. gevent.subprocess forks and execs in Python, so meanwhile that child holds a copy of every open TLS session: a copy writes a record, the parent reuses its sequence number, and the server answers bad_record_mac. Delete the register_at_fork line for the control.

```python
❯ uv run --script gevent_fork_ssl_poisoning_repro.py
gevent 26.7.0 | CPython 3.14.6
  File "src/gevent/greenlet.py", line 912, in gevent._gevent_cgreenlet.Greenlet.run
    result = self._run(*self.args, **self.kwargs)
  File "/home/guru/Desktop/gevent/gevent_fork_ssl_poisoning_repro.py", line 71, in talk
    sock.recv(4)
    ~~~~~~~~~^^^
  File "/home/guru/.cache/uv/environments-v2/gevent-fork-ssl-poisoning-repro-fef66dedca7c4d17/lib/python3.14/site-packages/gevent/ssl.py", line 615, in recv
    return self.read(buflen)
           ~~~~~~~~~^^^^^^^^
  File "/home/guru/.cache/uv/environments-v2/gevent-fork-ssl-poisoning-repro-fef66dedca7c4d17/lib/python3.14/site-packages/gevent/ssl.py", line 439, in read
    return self._sslobj.read(nbytes or 1024)
           ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^
ssl.SSLError: [SSL: SSLV3_ALERT_BAD_RECORD_MAC] sslv3 alert bad record mac (_ssl.c:2713)
2026-07-29T18:16:51Z <Greenlet at 0x74e02e604720: talk> failed with SSLError

POISONED: SSLError(1, '[SSL: SSLV3_ALERT_BAD_RECORD_MAC] sslv3 alert bad record mac (_ssl.c:2713)')

```

```python
# /// script
# requires-python = ">=3.10"
# dependencies = ["gevent==26.7.0", "greenlet==3.5.4", "trustme==1.2.1"]
# ///
from gevent import monkey; monkey.patch_all()   # first, before anything it patches

import sys

if sys.argv[1:2] == ['--server']:       # this file re-executed as the TLS peer, in a
    import socket                       # separate process, so no fork of ours reaches it
    import ssl
    ctx = ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)
    ctx.load_cert_chain(sys.argv[2])
    srv = socket.socket()
    srv.bind(('127.0.0.1', 0))
    srv.listen(1)
    print(srv.getsockname()[1], flush=True)
    conn = ctx.wrap_socket(srv.accept()[0], server_side=True)
    while True:
        try:
            data = conn.recv(4096)
        except OSError:                 # the MAC failure; the alert is already sent
            break
        if not data:
            break
        conn.sendall(data)
    sys.exit()


import os
import socket
import ssl
import subprocess

import gevent
import trustme

print('gevent %s | CPython %s'
      % (gevent.__version__, '.'.join(str(v) for v in sys.version_info[:3])))

ca = trustme.CA()
cert = ca.issue_cert('localhost')
with cert.private_key_and_cert_chain_pem.tempfile() as pem:
    server = subprocess.Popen([sys.executable, os.path.abspath(__file__), '--server', pem],
                              stdout=subprocess.PIPE)
    port = int(server.stdout.readline())

# The least a fork handler can do and still yield. Delete for the control run.

ctx = ssl.create_default_context()
ca.configure_trust(ctx)
sock = ctx.wrap_socket(socket.create_connection(('127.0.0.1', port)),
                       server_hostname='localhost')

def talk():
    while True:
        sock.sendall(b'ping')
        sock.recv(4)
        gevent.sleep(0.002)

talker = gevent.spawn(talk)
gevent.sleep(0.1)

for _ in range(20):
    subprocess.run([sys.executable, '-c', 'pass'], check=False)
    gevent.sleep(0.01)
    if talker.dead:
        break

server.kill()
if talker.dead:
    print('POISONED: %r' % (talker.exception,))
    sys.exit(1)
talker.kill()
print('TLS connection intact')

```

ddorian:
See failing tests:

https://github.com/ddorian/gevent/pull/3
https://github.com/ddorian/gevent/pull/4

Fixes #2196.

### DIFF
--- a/docs/changes/2196.bugfix.rst
+++ b/docs/changes/2196.bugfix.rst
@@ -1,0 +1,29 @@
+Give the child of :mod:`gevent.subprocess` a hub of its own, so that it runs
+none of the parent's greenlets before it ``exec``s.
+
+The POSIX ``_execute_child`` forks and execs in Python, so for that window the
+child is a whole interpreter holding a copy of the parent's greenlets, open
+file descriptors and in-memory TLS sessions. gevent's own child branch never
+switches, but ``os.register_at_fork(after_in_child=)`` handlers run inside
+``fork()`` before it begins; applications register them (OpenTelemetry, Sentry,
+``filelock``, ``coverage``), and under monkey-patching the ordinary contents of
+such a handler yield. The child's copy of the hub then resumed the copies,
+which carried on doing the application's work in a process that only existed to
+``exec``: writing on a pooled TLS connection until the peer answered
+``bad_record_mac``, or taking bytes off a pipe the parent was capturing, which
+has no MAC to notice it and so showed up only as a short or empty capture.
+
+The child now gets a fresh hub, which puts the parent's watchers, its queued
+callbacks and its hub greenlet out of reach. A fork handler that yields still
+works. One that waits for something only the parent could provide no longer
+runs application greenlets while it waits; it fails with
+:exc:`~gevent.exceptions.LoopExit` instead, unless it left work on the new loop
+first, in which case the child stays, with the parent blocked reading its error
+pipe inside ``Popen.__init__``.
+
+This also fixes the reproducer in :issue:`2023`, where a fork handler that used
+``gevent.get_hub().threadpool`` deadlocked the spawn.
+
+It applies only to the child of :mod:`gevent.subprocess`, which exists to
+``exec`` and nothing else. A child of a plain ``os.fork()`` may legitimately
+keep using the gevent objects it inherited, so it keeps the hub it was given.
--- a/src/gevent/subprocess.py
+++ b/src/gevent/subprocess.py
@@ -302,6 +302,103 @@ else:
     fork = monkey.get_original('os', 'fork')
     from gevent.os import fork_and_watch
 
+    from gevent.hub import Hub
+    from gevent.hub import set_hub
+    from gevent.hub import set_loop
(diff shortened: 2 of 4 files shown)


jamadden:
~I'm afraid this is a non-starter. It has long been documented behaviour that greenlets started in the parent continue to run in the child. It's too late to change that.~

Hang on, let me more closely consider ramifications...

ddorian:
Alternatives to this would be these:
https://github.com/gevent/gevent/pull/2203
https://github.com/gevent/gevent/pull/2205

jamadden:
Module-level state makes me nervous in a multi-threaded world. Can you address how this would work if two threads attempt to fork (and thus use `_forking_for_exec_in`) at the same time? Is there a solution that's inherently multi-thread safe?

ddorian:
I have since moved the project to free-threading so I no longer need this.

PR #2203, opened 2026-08-18
gevent/threading.py: cancel the parent's queued greenlet switches in a forked child

### LINKED ISSUE #2202: A forked gevent child runs copies of the parent's greenlets.
### Description:

The child does nothing but ``os._exit()``. The at-fork handler below yields, and that alone is enough for the child's hub to resume the parent's greenlets, inside ``fork()``, before it has returned.

### What I've run:

```python-traceback
❯ uv run repro_fork_child_queued_switches.py
10 of 10 children ran the parent greenlets: [117633, 117634, 117635, 117636, 117637, 117638, 117639, 117640, 117641, 117642]
```

```python
# /// script
# requires-python = ">=3.10"
# dependencies = ["gevent"]
# ///
from gevent import monkey; monkey.patch_all()

import os
import tempfile

import gevent

fileno, path = tempfile.mkstemp()
os.close(fileno)
fileno = os.open(path, os.O_WRONLY | os.O_APPEND)
parent = os.getpid()
stop = []

def worker():
    while not stop:
        os.write(fileno, b'%d\n' % os.getpid())
        gevent.sleep(0)

workers = [gevent.spawn(worker) for _ in range(4)]
gevent.sleep(0.05)
for _ in range(10):
    if os.fork() == 0:
        os._exit(0)
    os.wait()
    gevent.sleep(0.01)
stop.append(1)
gevent.joinall(workers)
os.close(fileno)

with open(path) as f:
    strangers = sorted({int(line) for line in f} - {parent})
os.unlink(path)

print('%d of 10 children ran the parent greenlets: %s' % (len(strangers), strangers))

```

ddorian:
Fixes #2202

### DIFF
--- a/docs/changes/2202.bugfix.rst
+++ b/docs/changes/2202.bugfix.rst
@@ -1,0 +1,11 @@
+Stop a forked child from running copies of the parent's greenlets.
+
+``_ForkHooks`` already made the copies appear stopped to :mod:`threading`, but
+every entry on the loop's callback queue was still a pending switch into one of
+them, and ``os.register_at_fork(after_in_child=)`` handlers yield under
+monkey-patching from inside ``fork()`` itself. A child that did nothing but
+``os._exit()`` ran four of the parent's greenlets on every one of 20 forks.
+
+Those switches are now cancelled. A greenlet parked on a timer or an ``io``
+watcher is resumed by the loop rather than by that queue, and this change does
+not reach it.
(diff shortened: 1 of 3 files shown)


jamadden:
I don't think we can do this. People deliberately start greenlets and have them doing things and then fork, expecting them to continue running. It has always been the case that those greenlets continue running in the child. This would badly break those scenarios --- we can't assume that every fork is going to be followed by an exec such that we don't care what greenlets are running.

PR #2205, opened 2026-08-18
[2204] Fix fork child pending waits 

### LINKED ISSUE #2204: A forked gevent child resumes greenlets that were blocked at the fork
### Description:

A runnable greenlet sits on the loop's callback queue; a blocked one is parked on a watcher, and the loop resumes it from there without consulting that queue. The workers here sleep between writes, so at any fork most of them are parked.

```python-traceback
❯ uv run repro_fork_child_pending_waits.py
17 of 20 children resumed the parent greenlets: [244157, 244162, 244163, 244164, 244166, 244167, 244168, 244169, 244170, 244172, 244173, 244174, 244176, 244178, 244179, 244180, 244181]
```

### What I've run:

```python
# /// script
# requires-python = ">=3.10"
# dependencies = ["gevent"]
# ///
from gevent import monkey; monkey.patch_all()

import os
import tempfile

import gevent

fileno, path = tempfile.mkstemp()
os.close(fileno)
fileno = os.open(path, os.O_WRONLY | os.O_APPEND)
parent = os.getpid()
stop = []

def worker():
    while not stop:
        gevent.sleep(0.002)  # parked on a timer, not sitting in the queue
        os.write(fileno, b'%d\n' % os.getpid())

workers = [gevent.spawn(worker) for _ in range(4)]
gevent.sleep(0.05)
for _ in range(20):
    pid = os.fork()
    if pid == 0:
        os._exit(0)
    while os.waitpid(pid, os.WNOHANG)[0] == 0:  # blocking waitpid can hang under libuv
        gevent.sleep(0.01)
    gevent.sleep(0.01)
stop.append(1)
gevent.joinall(workers)
os.close(fileno)

with open(path) as f:
    strangers = sorted({int(line) for line in f} - {parent})
os.unlink(path)

print('%d of 20 children resumed the parent greenlets: %s' % (len(strangers), strangers))

```

ddorian:
fixes #2204

### DIFF
--- a/docs/changes/2204.bugfix.rst
+++ b/docs/changes/2204.bugfix.rst
@@ -1,0 +1,15 @@
+Stop a forked child from resuming greenlets that were blocked at the fork.
+
+The companion to cancelling the queued switches: that queue only holds
+greenlets that were runnable, while a blocked one is parked on a watcher and
+the loop resumes it from there. In the child both are copies, and one resumed
+on an inherited TLS session writes a record under a sequence number the parent
+will use again, drawing ``bad_record_mac`` from the server.
+
+A watcher whose callback is a ``Waiter.switch`` is one a greenlet is blocked
+on; the ``active`` and ``pending`` flags cannot answer this, because libuv
+stops an expired timer before dispatching it. Only other greenlets' waits are
+cancelled, and only the callback, never the object, so a child of a plain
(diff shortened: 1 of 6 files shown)


jamadden:
As with the other PR that messes with generic forking, I don't believe we can accept this.

PR #2208, opened 2026-08-22
Avoid redundant hook setters in ThreadPool workers

### LINKED ISSUE #2206: Skip missing profiling and tracing hooks in ThreadPool workers
`_WorkerGreenlet` currently makes four setter calls for every task, even when
no profiling or tracing hooks are configured:

```python
def _before_run_task(...):
    sys.setprofile(_get_thread_profile())
    sys.settrace(_get_thread_trace())

def _after_run_task(...):
    sys.setprofile(None)
    sys.settrace(None)
```

Both getters return `None` unless a hook has been registered through the
`threading` module. For applications that do not use profiling or tracing,
this means every task is wrapped in four calls that simply set `None` to
`None`. We should skip those calls for three reasons:

1. **It matches the standard library.**
   [`threading.Thread._bootstrap_inner`][threading-bootstrap] only calls the
   `sys` setters when a hook has been registered:

   ```python
   if _trace_hook:
       _sys.settrace(_trace_hook)
   if _profile_hook:
       _sys.setprofile(_profile_hook)
   ```

2. **The setters have observable side effects.** `sys.setprofile()` and
   `sys.settrace()` each emit a PEP 578 audit event, even when passed `None`.
   An audit hook can reject the operation by raising an exception, so an
   application that never enabled profiling or tracing can still fail when
   gevent tries to disable hooks that were not enabled in the first place.

3. **The setters still make the interpreter do work.** On CPython 3.14, the
   legacy [profile setter][profile] and [trace setter][trace] each emit the
   audit event, stop the interpreter's threads, update the thread state and
   monitoring events, and restart the threads — even for a redundant
   `None`-to-`None` assignment. The cost may be small compared with dispatching
   a task, and any workload-level improvement should be measured rather than
   assumed. Still, it is unnecessary overhead, and the global coordination may
   matter more in a process with several active native threads.

When a hook *is* configured, the behavior stays the same: read the hooks just
before each task, install them for the task body, and remove them afterward.
We still need to read them before every task rather than once at worker
startup, because workers are long-lived and the hooks can change between
tasks.

## Implementation outline

```python
def _before_run_task(self, func, args, kwargs, thread_result, ...):
    profile = _get_thread_profile()
    trace = _get_thread_trace()

    self._task_profile = profile
    self._task_trace = trace

    if profile is not None:
        sys.setprofile(profile)
    if trace is not None:
        sys.settrace(trace)

def _after_run_task(self, func, args, kwargs, thread_result, ...):
    profile = self._task_profile
    trace = self._task_trace
    self._task_profile = None
    self._task_trace = None

    try:
        if profile is not None:
            sys.setprofile(None)
    finally:
        if trace is not None:
            sys.settrace(None)
```

The `try`/`finally` ensures that a failure while clearing the profile hook does
not leave the trace hook installed. These methods could use broader error
handling as well: a configured hook may still fail to install, and that path
must resolve the task's `ThreadResult` instead of leaving its caller stranded.
That is separate hardening work and can be reviewed independently.

bojanz:
ThreadPool workers called sys.setprofile() and sys.settrace() around every task, even when no hooks were configured. These calls do unnecessary thread coordination and emit PEP 578 audit events.

Only install configured hooks, matching threading.Thread.

Fixes #2206.

### DIFF
--- a/docs/changes/2206.bugfix.rst
+++ b/docs/changes/2206.bugfix.rst
@@ -1,0 +1,3 @@
+ThreadPool workers no longer call the profile and trace setters when no
+hooks are configured. This matches ``threading.Thread`` and avoids
+unnecessary PEP 578 audit events and thread coordination.
--- a/src/gevent/tests/test__threadpool.py
+++ b/src/gevent/tests/test__threadpool.py
@@ -819,6 +819,83 @@ class TestWorkerProfileAndTrace(TestCase):
     def test_trace_called_in_task(self):
         self._test_func_called_in_task('trace')
 
+    def _make_recording_pool(self):
+        class RecordingSys(object):
+            def __init__(self):
+                self.profile = None
+                self.trace = None
+                self.setter_calls = []
+
+            def getprofile(self):
+                return self.profile
+
+            def gettrace(self):
+                return self.trace
+
+            def setprofile(self, value):
+                self.setter_calls.append(('setprofile', value))
+                self.profile = value
+
+            def settrace(self, value):
+                self.setter_calls.append(('settrace', value))
+                self.trace = value
+
+        recording_sys = RecordingSys()
+
+        def no_hook():
+            return None
+
+        class Pool(ThreadPool):
+            class _WorkerGreenlet(ThreadPool._WorkerGreenlet):
+                # pylint:disable=signature-differs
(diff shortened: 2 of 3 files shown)


jamadden:
Thank you!

PR #2209, opened 2026-08-22
Resolve the ThreadResult when worker task setup fails

### LINKED ISSUE #2207: ThreadPool hangs indefinitely when `_before_run_task` fails
`_WorkerGreenlet.__run_task` calls `_before_run_task` outside the `try` block
responsible for resolving the `ThreadResult` (`src/gevent/threadpool.py`,
around line 170 in both 26.8.0 and master):

```python
    def __run_task(self, func, args, kwargs, thread_result):
        self._before_run_task(func, args, kwargs, thread_result)   # unguarded
        try:
            thread_result.set(func(*args, **kwargs))
        except: # pylint:disable=bare-except
            thread_result.handle_error((self, func), self._exc_info())
        finally:
            self._after_run_task(func, args, kwargs, thread_result)
            del func, args, kwargs, thread_result
```

If `_before_run_task` raises an exception, the worker logs "Failed to run
worker thread" and exits without setting the result. The greenlet waiting in
`ThreadPool.apply()` or `spawn().get()` then blocks forever, with no exception
or timeout.

This can happen in practice. `_before_run_task` calls `sys.setprofile()` and
`sys.settrace()`, both of which emit PEP 578 audit events. If an audit hook
rejects `"sys.setprofile"`, as sandboxing tools sometimes do, every task hangs
in this way.

Each stranded task also leaks a pool slot permanently. `spawn()` acquires the
worker semaphore and passes `semaphore.release` to the `ThreadResult`, which
only calls it when the result is resolved. The semaphore allows `maxsize + 1`
pending tasks. After that many failures, `spawn()` blocks for every caller,
even after the original cause has been removed. Because `FileObject` sends
`read`, `write`, and `close` operations through the hub's pool (whose default
`maxsize` is 10), eleven stranded tasks block every `FileObject` operation in
the process.

## Reproduction

[reproducer.py](https://gist.github.com/bojanz/b014eba60c42561969c1ccb6d80b2126) uses an audit hook to make the real `sys.setprofile()` call
fail, without replacing any gevent internals. The `gevent.Timeout` wrappers
are only there to let the script finish; without them, the calls never return.

```text
$ python3.14 reproducer.py
gevent 26.8.0 on python 3.14.6
baseline: returned 1
bug: HUNG - <ThreadPool at 0x... tasks=0 size=0 maxsize=10 hub=<Hub ...>>
slot leak, trigger removed: HUNG - <ThreadPool at 0x... tasks=0 size=0 maxsize=2 hub=<Hub ...>>
fixed: raised at the call site: sys.setprofile rejected by audit hook
fixed pool still works: returned 1
```

After the first hang, the pool reports `tasks=0 size=0`. In other words, the
broken pool looks idle, which makes the problem difficult to diagnose in a
running process.

## Suggested fix

The `ThreadResult` should be resolved even when setup fails, so the original
exception reaches the call site:

```python
    def __run_task(self, func, args, kwargs, thread_result):
        try:
            self._before_run_task(func, args, kwargs, thread_result)
        except BaseException:
            thread_result.handle_error((self, func), self._exc_info())
            raise
        try:
            thread_result.set(func(*args, **kwargs))
        ...
```

The reproducer confirms that this raises the exception from `apply()`, releases
the semaphore slot, and leaves the pool usable. Two additional hardening steps
would also help. First, install the profile and trace hooks only when they are
set, as `threading.Thread._bootstrap_inner` does; this removes both the failure
point and the per-task overhead. Second, have `run()` resolve the current
task's `ThreadResult` before allowing the worker to exit, which would protect
against future regressions along this path.

A regression test for the audit-hook case needs to run in a subprocess because
audit hooks cannot be removed. The behavior of `__run_task` itself can be
tested in-process, as shown in `reproducer.py`.

## Environment

- gevent 26.8.0 on CPython 3.14.6, Linux x86_64
- The same code is present on master at `402b8026`

bojanz:
__run_task called _before_run_task outside the block that resolves the ThreadResult. If it raised (for example, an audit hook rejecting sys.setprofile()), the caller of apply() or spawn().get() blocked forever and the pool slot was never released; after maxsize + 1 such failures every spawn() blocked too.

Deliver the exception through ThreadResult.handle_error before re-raising, so the caller gets it and the slot is released. The worker still exits and is replaced on demand.

Fixes #2207.


jamadden:
Thank you!

Unfortunately, we have a [consistent test failure](https://github.com/gevent/gevent/actions/runs/32583311892/job/104484075953) under 3.11.15/Ubuntu. This configuration runs leakchecks, meaning that each test is executed multiple times in the same process, so that test hygiene (ensuring the test leaves things in exactly the base state at the end of every run) is critical. You can probably reproduce this locally with the GEVENTTEST_LEAKCHECK=1 environment variable, and probably on any Python version. Can you take a look?

(There is a nuclear option of disabling a test when leakchecks are running, but it's really best not to do that unless there's no other option.)

bojanz:
Oops, sorry about that. Fixed the test and improved the comments. Also ported a fix for another corner case that we fixed while running this internally.

PR #2211, opened 2026-09-01
Report the full binary mode string from FileObjectPosix

### LINKED ISSUE #2039: Popen->FileObjectPosix has `w` mode while `wb` was expected
It seems there is an important difference in input streams of a subprocess. The mode is reported as `w` while `wb` was expected. This is based on Python 3 from Fedora 40, a virtualenv from it and gevent `24.2.1` installed via `pip install gevent`.

```python
#!/usr/bin/env python3

import gevent.monkey
gevent.monkey.patch_all(subprocess=True)

import subprocess
sdcat1 = subprocess.Popen(["systemd-cat", "-t", "python-test"], stdin=subprocess.PIPE, universal_newlines=True)
sdcat2 = subprocess.Popen(["systemd-cat", "-t", "python-test"], stdin=subprocess.PIPE)

print(sdcat1.stdin.mode)
print(sdcat2.stdin.mode)
print(type(sdcat2.stdin))
sdcat1.stdin.write("Hey")
sdcat2.stdin.write("Hey")
```

```
❯ python --version
Python 3.12.3
```

```
❯ python encoding2.py
w
w
<class 'gevent._fileobjectposix.FileObjectPosix'>
  File "encoding2.py", line 13, in <module>
    sdcat2.stdin.write("Hey")
TypeError: a bytes-like object is required, not 'str'
```

```
❯ pip freeze
gevent==24.2.1
greenlet==3.0.3
setuptools==70.0.0
zope.event==5.0
zope.interface==6.4.post2
```

Running it unpatched gives us `_io.TextIOWrapper` without `mode` attribute for `sdcat1` and `<class '_io.BufferedWriter'>` with `wb` `mode` for `sdcat2`.

afonsojanu:
Fixes #2039

`GreenFileDescriptorIO.mode` was being set from `open_descriptor.fileio_mode`, which is the mode string stripped down to just the read/write/append/create flag, meant for constructing the underlying `io.FileIO` object internally. That's not what should be surfaced to callers, since real `io.FileIO` includes the `b`/`t` suffix in its own `mode` attribute.

Under monkey-patched `subprocess.Popen`, this meant `process.stdin.mode` reported `'w'` even when the pipe was opened binary and only accepted bytes, which is exactly what the linked issue describes, and matches the `TypeError: a bytes-like object is required, not 'str'` from trying to `.write()` a str to it based on trusting `.mode`.

`OpenDescriptor` already computes the correct value (`fileio_mode` plus the `t`/`b` suffix) under its own `mode` attribute, this just uses that instead.

Added `test_mode_includes_binary_suffix` to `TestFileObjectPosix`, checking both the binary and text-mode-off case report `wb`/`rb` correctly. Confirmed it fails with the old `'w'`/`'r'` values by reverting just the source change, and passes with the fix. Ran the full `test__fileobject.py` (79 tests) and `test__subprocess.py` (36 tests) suites afterward with no other failures, and `pylint` on the touched source file scores 10.00/10.

### DIFF
--- a/src/gevent/_fileobjectposix.py
+++ b/src/gevent/_fileobjectposix.py
@@ -45,7 +45,11 @@ class GreenFileDescriptorIO(RawIOBase):
         self._closefd = closefd
         self._fileno = fileno
         self.name = fileno
-        self.mode = open_descriptor.fileio_mode
+        # Use the full mode (including the 't'/'b' suffix), matching what
+        # io.FileIO itself reports; open_descriptor.fileio_mode is the
+        # stripped-down string meant only for constructing the raw FileIO,
+        # not for user-facing introspection.
+        self.mode = open_descriptor.mode
         make_nonblocking(fileno)
         readable = open_descriptor.can_read
         writable = open_descriptor.can_write
--- a/src/gevent/tests/test__fileobject.py
+++ b/src/gevent/tests/test__fileobject.py
@@ -469,6 +469,19 @@ class TestFileObjectPosix(ConcurrentFileObjectMixin, # pylint:disable=too-many-a
     def _getTargetClass(self):
         return fileobject.FileObjectPosix
 
+    def test_mode_includes_binary_suffix(self):
+        # https://github.com/gevent/gevent/issues/2039
+        # Constructing with an explicit 'wb'/'rb' mode should report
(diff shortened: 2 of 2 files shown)


jamadden:
Thank you. I'll add the towncrier release fragment myself.

